PR #7, opened 2013-02-15
jQuery is required in non-AMD build for bootstrap.js

passy:
Generating an app with Bootstrap, but w/o require.js fails at the moment because jQuery is missing.

I put it into the plugins.js, but I'm not sure if it should be included even without bootstrap.js, because it's included in the components.js regardless.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -134,6 +134,12 @@ AppGenerator.prototype.writeIndex = function writeIndex() {
     '                <ul>'
   ];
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
It should be included separately since it is also require even though you don't want bootstrap.

passy:
Okay, I'll redo this.

passy:
2nd try. This time jQuery is in it's own module.

PR #9, opened 2013-02-15
Include main.js in non-AMD builds

passy:


sindresorhus:
Merge this with the jquery block and make the output main.js, make sure jquery is first.

passy:
Hm, so just merging both didn't really work because there was one uglify rule in the Gruntfile that would concat everything under `<%%= yeoman.app %>/scripts/*.js` into `main.js` and effectively overriding it.

Did that rule actually still do anything?

sindresorhus:
Hmm, I'm pretty sure it can be removed.

passy:
it doesn't seem that useful with the explicit processing through comments in place.

app/templates/Gruntfile.js
```diff
@@ -147,16 +147,6 @@ module.exports = function (grunt) {
                     mainConfigFile: 'app/scripts/main.js'
                 }
             }
-        },<% } else { %>
-        uglify: {
-            dist: {
-                files: {
-                    '<%%= yeoman.dist %>/scripts/main.js': [
-                        '.tmp/scripts/*.js',
```
sindresorhus: you have to keep this, it's used to concatenate compiled coffescript

passy:
While the main issue is discussed in #12 I think this is a reasonable work-around. CoffeeScript is now compiled to `coffee.js` in both the server and dist task and does no longer override `main.js`.

sindresorhus:
The coffee.js file needs to be referenced?

passy:
Should we reference it always or only if it isn't empty?

sindresorhus:
Always for now, so people don't complain, and rather do more with the real fix ;)

passy:
Ugh, the uglify task fails if there are no input files. Obviously, this couldn't happen before, because it included the normal JavaScript files as well. Any idea what to do about it?

passy:
I think this comes pretty close to the original idea. The non-AMD app now generates two build-blocks, one for for JavaScript and one for CoffeeScript and it behaves the same in build and server mode, treating files individually. To illustrate how CoffeeScript works with this, it generates a file. It's really obvious how to remove the CoffeeScript file, if you don't plan to use it. Do you think this is a good approach?

PR #11, opened 2013-02-16
Yeoman generator has incorrect path to sass-bootstrap

aboutaaron:
Error via Compass: `error styles/main.scss (Line 1: File to import not found or unreadable: sass-bootstrap/lib/bootstrap.`

Perhaps there's a better way to reference bootstrap, but here's my initial solution.


aboutaaron:
Ahh, realized I didn't compile with Grunt. My fault.

floer32:
Thanks for mentioning what your solution was, this helped me.

kanitw:
@aboutaaron  what do you mean "didn't compile with grunt"

PR #14, opened 2013-02-18
Delete an extra comma

nightire:


passy:
Thanks, @nightire. This is already been taking care of in #9 among other things. :)

PR #19, opened 2013-02-22
grunt-rev

sebdeckers:
The build process should use grunt-rev to rename all files according to their content hash.

Note: Please check how this affects the "watch" task targets. I'm not sure if rev/prepareUsemin/usemin need to be called each time now, or if the existing logic continues to work correctly.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -171,9 +171,17 @@ module.exports = function (grunt) {
             }
         },<% } %>
         rev: {
-            dist: {
+            src: {
                 files: {
-                    src: ['.tmp/**/*.{js,css,png,jpg}']
+                    src: ['.tmp/**/*.{js,css}']
+                }
+            },
+            assets: {
+                files: {
+                    src: [
+                        '<%%= yeoman.app %>/images/**/*.{png,jpg,jpeg}',
+                        '<%%= yeoman.app %>/styles/fonts/**/*.*'
+                    ]
                 }
             }
         },
@@ -283,10 +291,11 @@ module.exports = function (grunt) {
         'test',
         'coffee',
         'compass:dist',
-        'rev',
+        'rev:src',
         'useminPrepare',<% if (includeRequireJS) { %>
         'requirejs',<% } %>
         'imagemin',
+        'rev:assets',
         'cssmin',
         'htmlmin',
         'concat',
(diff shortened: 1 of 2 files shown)


sebdeckers:
Also see: yeoman/grunt-usemin#56

sebdeckers:
cc @sleeper @sindresorhus

app/templates/Gruntfile.js
```diff
@@ -170,6 +170,13 @@ module.exports = function (grunt) {
                 }
             }
         },<% } %>
+        rev: {
+            dist: {
+                files: {
+                    src: ['.tmp/**/*.{js,css,png,jpg}']
```
sindresorhus: You need to add in `.jpeg` too

passy: `.webp`? :)

sindresorhus: Sure that too.

sindresorhus: The config is not correct though, since images will not be in .tmp

sebdeckers: It looks like I'll have to split this into two targets. One for html/js and anotehr for images. Because images should only be revved once they have been copied with the imagemin task. Alternatively, I should implement file moving (dest) in the rev task, to drop the revved file names into .tmp, where imagemin can pick them up to move to dist. Thoughts?

sindresorhus: I think the task should have file moving. Old yeoman didn't do that for some reason, but it's standard grunt practice.

app/templates/Gruntfile.js
```diff
@@ -170,6 +170,21 @@ module.exports = function (grunt) {
...
+        rev: {
+            src: {
+                files: {
+                    src: ['.tmp/**/*.{js,css}']
+                }
+            },
+            assets: {
+                files: {
+                    src: [
+                        '<%%= yeoman.app %>/images/**/*.{png,jpg,jpeg}',
```
sindresorhus, line 182: webp

sindresorhus:
> It's not clear to me when/where fonts are being copied to the dist folder. Is this correct?

Yes

> Also, should I run rev on favicon.ico?

No. Server logs go crazy when favicon.ico is not in the default location.

sindresorhus:
Also fix merge conflicts

sindresorhus:
ping

sindresorhus:
ping. Can you also fix the merge conflict?

passy:
Continued in #33.

PR #20, opened 2013-02-22
Add support for globbing one level down

sindresorhus:
Initially I limited it to only the specified folder, because of performance problems we experienced in 0.9.x.

There has come up a lot of questions about our globbing. Especially about the use-case of having eg. images in separate folders `app/images/ui/` `app/images/content`.

This proposal adds the ability to glob one level down. Which means it would match every image in `app/images` and `app/images/ui`, but not `app/images/ui/anotherdir`.

What do you think? Can you also look through the PR and see if it's ok.
## 

@sleeper @btford @passy @addyosmani

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -4,6 +4,12 @@ var mountFolder = function (connect, dir) {
     return connect.static(require('path').resolve(dir));
 };
 
+// # Globbing
+// for performance reasons we're only matching one level down:
+// 'test/spec/{,*/}*.js'
+// use this if you want to match all subfolders:
+// 'test/spec/**/*.js'
+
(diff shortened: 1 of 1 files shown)


passy:
Looks good to me for now, but I think in the long run we should look into alternative mechanisms for watching. Even if the limit for open files per process isn't exceeded, the polling is quite costly CPU-wise. Watching a small-sized project with a couple of dozen sass and coffee files keeps the fans on my notebook constantly spinning.

sindresorhus:
@passy these might help:

https://npmjs.org/package/inotify-plusplus
https://npmjs.org/package/watch-inotify

btford:
:+1: 

I just mirrored these changes in `generator-angular`.

PR #23, opened 2013-02-25
Copying scripts/*.js files when requirejs is activated

fcamblor:
Will allow to require 'lately' scripts not concatenated by requirejs task, see [this issue](https://github.com/yeoman/yeoman/issues/934) and [this repo demonstrating the issue](https://github.com/fcamblor/yeoman-requirejs-grunt-build)

Note that this copy is only needed when flag `includeRequireJS` is true, otherwise, js files are all concatenated into main.js (see `uglify` task in that case).

I would like to filter vendor/bootstrap.js too, when sass for bootstrap is activated, but I didn't found the flag name used to test this case. If you can help me on this flag name, I can edit this PR to handle this case.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -235,7 +235,10 @@ module.exports = function (grunt) {
                     dest: '<%%= yeoman.dist %>',
                     src: [
                         '*.{ico,txt}',
-                        '.htaccess'
+                        '.htaccess'<% if (includeRequireJS) { %>,
+                        'scripts/**/*.js',
+                        '!scripts/main.js',
+                        '!scripts/vendor/require.js'<% } %>
                     ]
                 }]
             }


sindresorhus:
// @passy @btford any thoughts?

passy:
Haven't thought about this case before. It certainly makes sense to support lazy loading, since this is one of the biggest benefit of using AMD after all. However, the exceptions are hard to spot and could lead to some unexpected behavior for newcomers. That said, I don't have any better idea at the moment.

sindresorhus:
I don't have much experience with RJS. Would be nice to get someone else that does to comment on this.

Another issue with doing this by default is that we copy over all these script files which most probably won't use. And then they'll complain there's a lot of deadweight in the dist folder, like they did before and why we only whitelist files now.

passy:
That's a valid concern. If we leave it like this, what would be a good place to document this?

sindresorhus:
@passy since this probably applies to more than this generator I think we should have it in the global wiki. Where in the wiki you say. Well, not sure. Either as a RJS page or something or in the FAQ or a tips and tricks page?

## 

@kevva do you have any experience with RJS or thoughts about how we should handle lazy loading?

passy:
@sindresorhus RJS questions come up all the time, so I guess having a page dedicated to RJS/AMD topics could be useful.

fcamblor:
:+1: for the wiki entry, at least :)

addyosmani:
I actually think this would make for a useful addition. Would require rebasing for us to land it however.

sindresorhus:
@addyosmani see my comments, there are some unresolved concern about copying over all the files even if they're not needed.

rikkert:
I use RJS on my projects and always end up concatenating to one file.
It would be confusing if I have a dist folder with all the JS files and I would also complain about the deadweight. ;)

Maybe we can do it as an option? Even a question during generating is not bad I think.
If we would mention it on the wiki showing what to change in the Gruntfile can even be enough imo.

kevva:
Hey, haven't payed attention to this before. Instead of copying over files, use `findNestedDependencies: true` in the `requirejs` task. This will concat nested `require()` calls.

If you're project is really big, it makes sense to divide your files into layers. See [this](http://requirejs.org/docs/1.0/docs/faq-optimization.html#priority). You shouldn't copy over files, that's for sure.

rikkert:
We should probably add that option in the default setup to make it at least work.
It does of course defeats the purpose of dynamic loading of certain scripts.
In some cases you do really want to use this option. Maybe only for a small part of the project.

sindresorhus:
I don't see this as something most will want and it should be easy to add it in for those who will. Would love a wiki page describing how to do this in the yeoman wiki: https://github.com/yeoman/yeoman/wiki

fcamblor:
What's the point in using RequireJS if this is not for dynamic loading scripts ?

I don't catch it.

sindresorhus:
My comment above is about this specific solution and not dynamic loading in general. Feel free to open a new ticket over at yeoman/yeoman with some good arguments. Maybe by having a discussion about it we can come up with a good solution together.

PR #27, opened 2013-02-28
Updated USAGE

passy:
I updated the USAGE based on the [Angular update](https://github.com/yeoman/generator-angular/blob/a296aeca1cbe669a30e89dc2f5ebdecec4a3b62d/app/USAGE), but added a few explanations. Wdyt?

### DIFF
--- a/app/USAGE
+++ b/app/USAGE
@@ -1,8 +1,17 @@
 Description:
-    The 'yeoman init' command creates a new web application.
+    Creates a new basic front-end web application.
+
+Options:
+    Twitter Bootstrap: Include Twitter Bootstrap for Sass
+    RequireJS: Add support for AMD-loading via RequireJS
 
 Example:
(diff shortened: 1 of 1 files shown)


app/USAGE
```diff
@@ -1,8 +1,17 @@
...
+
+    This will create:
+        Gruntfile.js: Configuration for the task runner.
+        component.json: Front-end packages installed by bower.
+        package.json: Development packages installed by npm.
 
-    This generates a skeleton for a web app.
-    See the README in the newly created application to get going.
+        app/: Your application files.
+        test/: Mocha-based unit tests.
```
sindresorhus: Just `unit tests`, we support Jasmine too

passy: Right, I forgot.

sindresorhus:
Thanks

PR #32, opened 2013-03-17
Auto install dependencies

sindresorhus:
Many users have voiced concern of there being too many commands to get started. The original intention of not automatically installing deps after scaffolding was for clearity of what was happening. Since users would think a npm bug was a bug with Yeoman. Seeing as this has helped nothing, and people keep opening invalid issues like crazy, I don't think that is an argument anymore.

After thinking about this for a while, I've come to the conclusion that we can do both clearity and ease of use, by doing it automatically, but being totally clear about it. That means showing a clear message as shown below and documenting it. After all, Yeoman is about making front-end development less painful.

@addyosmani @sleeper @btford @passy wdyt?

_this also parallizes both commands, so it's faster than before._

Screen Shot 2013-03-17 at 16 04 58

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,6 +1,7 @@
 'use strict';
 var util = require('util');
 var path = require('path');
+var spawn = require('child_process').spawn;
 var yeoman = require('yeoman-generator');
 
 
@@ -25,7 +26,9 @@ function AppGenerator(args, options, config) {
   this.mainCoffeeFile = 'console.log "\'Allo from CoffeeScript!"';
 
   this.on('end', function () {
-    console.log('\nI\'m all done. Just run ' + 'npm install && bower install'.bold.yellow + ' to install the required dependencies.');
+    console.log('\n\nI\'m all done. Running ' + 'npm install & bower install'.bold.yellow + ' for you to install the required dependencies. If this fails, try running the command yourself.\n\n');
+    spawn('npm', ['install'], { stdio: 'inherit' });
+    spawn('bower', ['install'], { stdio: 'inherit' });
   });
 
   this.pkg = JSON.parse(this.readFileAsString(path.join(__dirname, '../package.json')));


passy:
I've been thinking about this myself and had the same concerns that users could presume errors in the process would be caused by yeoman. I think this is a very elegant solution to that problem.

app/index.js
```diff
@@ -25,7 +26,9 @@ function AppGenerator(args, options, config) {
   this.mainCoffeeFile = 'console.log "\'Allo from CoffeeScript!"';
 
   this.on('end', function () {
-    console.log('\nI\'m all done. Just run ' + 'npm install && bower install'.bold.yellow + ' to install the required dependencies.');
+    console.log('\n\nI\'m all done. Running ' + 'npm install &; bower install'.bold.yellow + ' for you to install the required dependencies. If this fails, try running the command yourself.\n\n');
```
passy: Should be just `npm install & bower install`. The semicolon there is a syntax error, if I'm not mistaken.

sindresorhus: You were the one that suggested that syntax.

passy: I know. :persevere: Searching for the other occurrences right now.

passy: It works in fish, but not in bash.

sindresorhus: whatever, `npm install & bower install` is one char shorter, so why not. Also Bash sucks.

sindresorhus:
the ping machine pings @addyosmani @sleeper @btford ping pong pang

## 

// @paulirish @kevva

btford:
I like it. Make it so!

kevva:
Pong, looks good! One could argue that the error output by npm and bower should be enough to make a distinction between those and yeoman. Also, parallelizing is a big plus.

sindresorhus:
@addyosmani @paulirish last chance to object ;)

btford:
No objections. Do it.

addyosmani:
I'm 100% behind this change. We should definitely do it.

addyosmani:
PS: We need to make sure that if/when this lands we update all of the installation guides for relevant generators as the CLI will be providing instructions for the two steps we're aiming to automate here.

passy:
Yeah! :space_invader:

sindresorhus:
@addyosmani ticket: https://github.com/yeoman/yeoman/issues/1007

yveslaroche:
I get the following error after running "yo webapp" now on Windows using the Git Bash. 

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: spawn ENOENT
    at errnoException (child_process.js:977:11)
    at Process.ChildProcess._handle.onexit (child_process.js:768:34)
```

However, running "npm install & bower install" manually works just fine.

sindresorhus:
@yveslaroche can you gist your full terminal output? what Node version?

@passy @kevva can you test?

yveslaroche:
I'm using node.js v0.10.3. My full terminal output is here: https://gist.github.com/yveslaroche/5337427

I've also linked in the "generator" repo with the "generator-webapp", as it is a dependency there is well. This step is missing from the contributing instructions. I noticed this because I had the following error before, which was fixed with the latest commit to the generator.

```
D:\Dropbox\Dev\GitHub\yeoman\generator-webapp\node_modules\yeoman-generator\node
_modules\bower\node_modules\tmp\lib\tmp.js:219
    throw err;
          ^
Error: spawn ENOENT
    at errnoException (child_process.js:977:11)
    at Process.ChildProcess._handle.onexit (child_process.js:768:34)
```

kevva:
@yveslaroche, try clearing your bower cache and make sure you've got the latest version of bower. May be related to twitter/bower#307.

yveslaroche:
@kevva I've cleared bower cache before. And just tried it again with bower v0.8.6 and the latest bower master branch, too. I'm still getting the same error https://gist.github.com/yveslaroche/5337427 with both bower versions.

addyosmani:
hmm. Haven't been able to reproduce. Anyone else have better luck?

PR #33, opened 2013-03-17
grunt-rev integration (updated)

passy:
I took @cbas' PR, squashed the commits and updated it to match the new globbing schema.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -187,6 +187,21 @@ module.exports = function (grunt) {
         /*uglify: {
             dist: {}
         },*/<% } %>
+        rev: {
+            src: {
+                files: {
+                    src: ['.tmp/**/*.{js,css}']
+                }
+            },
+            assets: {
+                files: {
+                    src: [
+                        '<%%= yeoman.app %>/images/**/*.{png,jpg,jpeg}',
+                        '<%%= yeoman.app %>/styles/fonts/**/*.*'
+                    ]
+                }
+            }
+        },
         useminPrepare: {
             html: '<%%= yeoman.app %>/index.html',
             options: {
@@ -308,9 +323,11 @@ module.exports = function (grunt) {
         'clean:dist',
         'coffee',
         'compass:dist',
+        'rev:src',
         'useminPrepare',<% if (includeRequireJS) { %>
         'requirejs',<% } %>
         'imagemin',
+        'rev:assets',
         'svgmin',
         'htmlmin',
         'concat',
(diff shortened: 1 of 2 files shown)


sebdeckers:
@passy Thanks!

passy:
@cbas You're welcome!

I just noticed that this breaks CoffeeScript support, because the compiled `.coffee` files gets revved in `.tmp/scripts/` and is subsequently not found by usemin. We need to find a solution before we merge this.

sebdeckers:
@passy Currently `grunt-rev` simply renames the file in place. It doesn't copy them to the destination path. Something to fix in `grunt-rev`, not `generator-webapp`, I think.

kevva:
I would perform the `rev` task on the assets in `dist` and not in `app` or `.tmp`. The `rev` task also has to be run after `uglify` `cssmin` etc, basically at the end before `usemin`.

Works for me. See kevva/generator-webapp@cdc76b3.

passy:
Makes sense. I'm going to upgrade the PR to that.

kevva:
You need to fix the build task too with the right targets.

passy:
After `uglify`, right?

kevva:
After `copy`, I think. The fonts, `.webp` and `.gif` files won't get revved now since they are copied over. On the other hand, I don't know if that will work anyway. Have to test that.

passy:
Right. I'm going to do some manual testing tonight to see if we didn't miss anything.

passy:
Alright, this seems to work as expected. However, this breaks any reference to resources from any place but HTML files, e. g. `url()` directives within CSS. How are those supposed to be used?

kevva:
Strange. I thought the usemin task handled those.

@sleeper @cbas

passy:
@kevva It does, I guess it must have been a caching issue. If you depend on a fixed path in JS you can still disable it easily enough. That should cover everything then or can you think of another edge case?

kevva:
> If you depend on a fixed path in JS you can still disable it easily enough. 

Yes, I believe there's an issue for JS on the usemin repo too.

> That should cover everything then or can you think of another edge case?

Yeah, that should do it.

sebdeckers:
The `usemin` task does support cache manifest nor source maps, afaik. It works reasonably well with most CSS and HTML references. You also want to run `rev` after RequireJS Optimizer.

passy:
@sindresorhus Want to take another look? lgtm

sindresorhus:
@passy lgtm. feel free to merge when the merge conflict is fixed. also needs committed to the angular generator.

PR #39, opened 2013-03-25
Uppercase html DOCTYPE

alefteris:
To be consistent with h5bp.


passy:
Why?

kevva:
WHY? I LIKE IT BETTER NOW. h5bp was lowercase too since not so long ago h5bp/html5-boilerplate@7a2f9dc.

alefteris:
To keep the delta with h5bp as small as possible? And what @mathiasbynens comments on the above mentioned commit?  Anyway, it's not a big deal :) Close if you like it as it is...

PR #49, opened 2013-04-10
Fix running `bower install` and `npm install` on Windows

kevva:
`child_process.spawn` currently only supports `.exe` files on Windows. Any better solution?


passy:
I prefer this over adding another dependency just for the workaround. I wonder if we should move the logic to `generator` in another step, though.

orestes:
Works for me on Win 7 x64. I've patched my generator-angular as well. Thanks @kevva

sindresorhus:
Not that I can think of. If anyone this. Please chime in.

Thanks @kevva

PR #58, opened 2013-04-29
New prompts for confirmations

passy:
I'd appreciate it if someone could try these locally with the unreleased `yeoman-generator` module. It works fine for me now, but I had an occasion on one machine where the cursor would be oddly misplaced when the prompts came up.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -53,13 +53,13 @@ AppGenerator.prototype.askFor = function askFor() {
   var prompts = [{
     name: 'compassBootstrap',
     message: 'Would you like to include Twitter Bootstrap for Sass?',
-    default: 'Y/n',
+    default: true,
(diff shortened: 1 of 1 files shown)


addyosmani:
Nice work! Are there any specific platforms you were experiencing that bug on? Happy to try reproducing if so.

passy:
It happened on Linux in a gnome-terminal in a tmux session. Haven't seen it again since then. It could have been just some weird interplay between some partly upgraded generators, but it would be nice to have some more people trying it.

passy:
Works fine in the Secure Shell extension for Chrome. :)

sindresorhus:
Works fine for me

passy:
Thanks for testing, guys!

PR #59, opened 2013-05-04
Use grunt-contrib-watch for livereload

sindresorhus:
https://github.com/gruntjs/grunt-contrib-watch#live-reloading

grunt-contrib-watch got support for livereload. We should convert all generators over to it asap.

<del>This is just a start as I was stopped by a potential [bug](https://github.com/intesso/connect-livereload/issues/1) in connect-livereload. @kevva @passy would you be able to look into it?</del>

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -1,6 +1,7 @@
 // Generated on <%= (new Date).toISOString().split('T')[0] %> using <%= pkg.name %> <%= pkg.version %>
 'use strict';
-var lrSnippet = require('grunt-contrib-livereload/lib/utils').livereloadSnippet;
+var LIVERELOAD_PORT = 35729;
+var lrSnippet = require('connect-livereload')({port: LIVERELOAD_PORT});
 var mountFolder = function (connect, dir) {
     return connect.static(require('path').resolve(dir));
 };
@@ -24,6 +25,9 @@ module.exports = function (grunt) {
     grunt.initConfig({
         yeoman: yeomanConfig,
         watch: {
+            options: {
+              spawn: false
+            },
             coffee: {
                 files: ['<%%= yeoman.app %>/scripts/{,*/}*.coffee'],
                 tasks: ['coffee:dist']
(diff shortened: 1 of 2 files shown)


sindresorhus:
There's also a problem of grunt-contrib-watch going into an infinite loop when `nospawn: true` is not used.

shama:
@sindresorhus What about just including the snippet in the `app/templates/index.html` with a comment?

sindresorhus:
@shama I have suggested that before and people didn't seem to like it. We could of course use some kind of HTML conditional, but we require the server for Compass/Coffee compiling anyway, so why not.

shama:
@sindresorhus Ah I see. It seems similar to including the existing modernizr or google analytics script. Just a helpful default the user could erase if they don't need it.

sindresorhus:
Yes, but then we would have a mechanism to remove it for production, which seems like more work than just adding a middleware to the connect task we're using anyway.

revathskumar:
@sindresorhus / @passy Is it okey to make these changes over [yeoman/generator-backbone](https://github.com/yeoman/generator-backbone)

sindresorhus:
Yes, go for it. I was going to mention you all but forgot.

sindresorhus:
@inkredabull @passy @PaulKinlan @ragingwind can you update your generators accordingly?

Don't use this PR though, as it's not entirely correct. See changes here: https://github.com/yeoman/generator-webapp/compare/v0.1.7...v0.2.2

inkredabull:
Sure, perhaps not until after I/O though ;-)

ragingwind:
I have no my laptop at this moment. Because i'm visit to google io in San Francisco for this week.  I will update asap after back to the home. Sorry for that

jeef3:
Maybe just me, but does this break live reloading (on Windows)?

addyosmani:
@sindresorhus does the Chrome app one have live-reload? I filed a PR for ember generator and briefly looked at chrome app but didn't see as much config in there for live reload as the others. Will look again later.

sindresorhus:
@addyosmani don't think so

cetchells:
@jeef3 Yes livereload  seems broken on win8 for me, Chrome doesn't refresh. The watch task runs ok, the socket.io connections seems to be active, just no refresh in the browser

ragingwind:
@sindresorhus @addyosmani @PaulKinlan current chrome extension / app generator is not using (i removed). after manifest v2, application's entry-point has been changed to background.js, not html. but some particular page can be use, popup/option and some legacy packaged-apps. i couldn't finding out how to use with live-reload. i must be used to chrome developer mode for test / run. any idea?

sindresorhus:
@ragingwind i don't think livereload makes sense for those, but can you still implement the other changes from the diff link in my previous comment?

ragingwind:
@sindresorhus sorry, i've misunderstood. i'll start now.

PR #66, opened 2013-05-18
Revise confirmation messages for new prompt

robwierzbowski:
With the read library, `warning` was a message based on successful user input. With the new prompt it seems to be an alias of prompt's `message`, which is an explanation of why your input didn't pass validation. 

This switches the confirmation messages to normal `console.log`s.

Shout out to #58.


robwierzbowski:
Wrote this before I learned that the prompt library was still being decided on. With https://github.com/yeoman/generator/pull/250 this isn't needed any longer.

PR #71, opened 2013-05-25
Added support for autoprefixer

### LINKED ISSUE #62: Autoprefixer
I think it could be a good selling point if we included this by default: https://github.com/nDmitry/grunt-autoprefixer

> Yeoman will also autoprefix your CSS so you don't have to.

@addyosmani @paulirish

passy:
Before I lose track of it, I pulled out my autoprefixer setup for #62.

Currently, this only works for files processed by compass, because it only runs on files in `.tmp`. In order to make sure files aren't processed twice by autoprefxier, we would have to restructure the tasks a little to make it work for plain CSS files. OTOH, I would argue if you use plain CSS, you might not want any magic at all.

But let's discuss those details post-1.0. :)

(fix #62)

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -38,15 +38,20 @@ module.exports = function (grunt) {
             },
             compass: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
-                tasks: ['compass:server']
-            },
+                tasks: ['compass:server'<% if (autoprefixer) { %>, 'autoprefixer' <% } %>]
+            },<% if (autoprefixer) { %>
+            styles: {
+                files: ['<%%= yeoman.app %>/styles/{,*/}*.css'],
+                tasks: ['copy:styles', 'autoprefixer']
+            },<% } %>
             livereload: {
                 options: {
                     livereload: LIVERELOAD_PORT
                 },
                 files: [
-                    '<%%= yeoman.app %>/*.html',
-                    '{.tmp,<%%= yeoman.app %>}/styles/{,*/}*.css',
+                    '<%%= yeoman.app %>/*.html',<% if (autoprefixer) { %>
+                    '.tmp/styles/{,*/}*.css',<% } else { %>
+                    '{.tmp,<%%= yeoman.app %>}/styles/{,*/}*.css',<% } %>
                     '{.tmp,<%%= yeoman.app %>}/scripts/{,*/}*.js',
                     '<%%= yeoman.app %>/images/{,*/}*.{png,jpg,jpeg,gif,webp,svg}'
                 ]
@@ -173,7 +178,20 @@ module.exports = function (grunt) {
                     debugInfo: true
                 }
             }
-        },
+        },<% if (autoprefixer) { %>
+        autoprefixer: {
+            options: {
+                browsers: ['last 1 version']
+            },
+            dist: {
+                files: [{
+                    expand: true,
+                    cwd: '.tmp/styles/',
+                    src: '{,*/}*.css',
+                    dest: '.tmp/styles/'
+                }]
+            }
+        },<% } %>
         // not used since Uglify task does concat,
         // but still available if needed
         /*concat: {
@@ -302,20 +320,30 @@ module.exports = function (grunt) {
                         'generated/*'
                     ]
                 }]
-            }
+            }<% if (autoprefixer) { %>,
+            styles: {
+                expand: true,
+                dot: true,
+                cwd: '<%%= yeoman.app %>/styles',
+                dest: '.tmp/styles/',
+                src: '{,*/}*.css'
+            }<% } %>
         },
         concurrent: {
             server: [
                 'coffee:dist',
-                'compass:server'
+                'copy:styles'<% if (autoprefixer) { %>,
+                'autoprefixer'<% } %>
             ],
             test: [
                 'coffee',
-                'compass'
+                'copy:styles'<% if (autoprefixer) { %>,
+                'autoprefixer'<% } %>
             ],
             dist: [
                 'coffee',
-                'compass:dist',
+                'compass',<% if (autoprefixer) { %>
+                'copy:styles',<% } %>
                 'imagemin',
                 'svgmin',
                 'htmlmin'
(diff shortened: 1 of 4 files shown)


app/templates/Gruntfile.js
```diff
@@ -173,6 +173,19 @@ module.exports = function (grunt) {
                 }
             }
         },
+        autoprefixer: {
+            options: {
+                browsers: ['last 2 versions']
```
sindresorhus: I would prefer that to be `last 1 version`

// @addyosmani

sindresorhus:
I would argue it's just as useful for vanilla CSS. I really dread having to manually add prefixes when doing vanilla CSS.

passy:
True, but you can still write vanilla CSS using SCSS. Having autoprefixer
only for otherwise preprocessed files would reduce the potential surprise
when the dist files don't match those you wrote previously. But I don't
have a strong opinion about this.

robwierzbowski:
> you can still write vanilla CSS using SCSS.

Although if a person isn't using a preprocessor it probably won't occur to them to preprocess their vanilla css. If surprise is a big consideration, autoprefixer could be added as a choice on the prompt.

I'm wondering if a second levels of css preprocessing is going to noticeably slow livereload, especially if livereload is targeting the watched css. It looks like livereload will fire once when compass creates a css file in .tmp, and then again when autoprefixer passes over it.

addyosmani:
I vote for a prompt. This is one of those magical things users should have a choice about :) I feel like doing this by default might end up causing issues with a workflow where you don't realize prefixing is being handled for you in the background.

passy:
I like the idea of a prompt, too. Regarding compilation speed, it's a negligible hit on the total performance as compass is slow as hell. `autoprefixer` itself is really fast.

sindresorhus:
I'm not a big fan of having a prompt for this. Prefixing should just work. But since the lib isn't fully tried out yet. I'd say we go for a prompt for now and can evaluate removing it at some point in the future.

robwierzbowski:
Note for vendor + preprocessor autoprefixing: 

If you want autoprefixer to process vendor css files to `.tmp`, and process Compass/css preprocessor result files in place in `.tmp`, vendor files become concatenated with themselves each time watch fires. Fixable with a css preprocessor intermediate directory or by a copy task that transfers all vendor css to tmp first, like the build process. I'm slightly in favor of the latter.

addyosmani:
@passy would you like to refactor this pull to include changes to bring this behind a prompt?

passy:
@addyosmani Sure! Do we still want to wait for the 1.0 release?

addyosmani:
I think we could aim for 1.0 RC by the end of next week or the week after. If you think you'd be able to implement and test this accordingly before then, happy to consider landing this pre RC. Because it's behind a prompt and still optional it's safer.

passy:
@addyosmani Alright, I should find time tomorrow night to get this done.

passy:
With support for plain CSS this became a pretty big change. I tested all combinations I could think of and everything seems to work nicely, but I would really appreciate if you could give this some testing love.

// @sindresorhus @addyosmani @stephenplusplus

robwierzbowski:
I think you can get away without the extra tmp/rawstyles directory, and a few less tasks.

**copy:stageCSS**
app/style/{,_/}_.css -> .tmp/style

**compass**
app/style -> .tmp/style

**autoprefixer**
.tmp/style -> .tmp/style

**watch:stageCSS**
files: app/styles/{,_/}_.css
task: copy:stageCSS

**watch:autoprefixer**
files: tmp/styles/{,_/}_.css
task: autoprefixer

Here's my [Gruntfile](https://github.com/robwierzbowski/generator-jekyllrb/blob/master/app/templates/Gruntfile.js) if you're curious. Is there something I'm missing that necessitates rawstyles?

passy:
Thanks, @robwierzbowski. I'm trying this now. This went through so many iterations, I'm not exactly sure what the initial idea behind this was.

robwierzbowski:
Yeah, I did too. This was the cleanest way I could find for server. By the way you were right: autoprefixer is super fast, no impact on observable livereload speed at all.

passy:
Updated to remove the `rawstyles` folder. Haven't squashed yet in case something comes up I haven't thought about.

robwierzbowski:
Nice, one less watch task even.

addyosmani:
Nice work!

On Thursday, May 30, 2013, Rob Wierzbowski wrote:

> Nice, one less watch task even.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/yeoman/generator-webapp/pull/71#issuecomment-18712771
> .

## 

Sent from Gmail Mobile

passy:
Thanks, @robwierzbowski. Copying is now within the concurrent tasks.

necolas:
> you can still write vanilla CSS using SCSS

But you can't write vanilla CSS with SCSS if you need to support prefixed implementations too. That's where an approach like that of Autoprefixer comes in. I'd much rather write "CSS of the future" and have the build step fill in the blanks.

passy:
> I'd much rather write "CSS of the future" and have the build step fill in the blanks.

I think that's a great way to sell it. We should use a phrase like that to promote it. Btw. this PR has support for running autoprefixer on vanilla `.css` now, too.

robwierzbowski:
Not sure what you mean by "if you need to support prefixed implementations too", but this task now adds autoprefixer support if you're using vanilla or preprocessed css. The "vanilla CSS using SCSS" comment was only because in the original implementation of the task used Compass to move files to the tmp directory where they were processed by autoprefixer.

robwierzbowski:
@passy, simulpost!

sindresorhus:
Is this ready to be merged?

You need to fix merge conflict.

passy:
I pushed an update, but I'm current on public park wifi :deciduous_tree: :wavy_dash:, which is probably by an order of magnitude worse than conference wifi, so I can't test my changes. If someone else wants to give it a try, feel free to do so. Otherwise, I'll check this out in the afternoon.

addyosmani:
@passy if you get a chance to test this out further that would be appreciated. We can also do a review during/after the meeting if you would like.

passy:
I just rebased it on top of master and haven't noticed any problems trying vanilla css / compass, live reload and the build step.

addyosmani:
Thanks for the update, @passy! Let's test this out some more this afternoon and if all is well we'll merge pre-RC1.

passy:
sgtm :+1:

passy:
@addyosmani Did you have a chance to try this out?

addyosmani:
Works well for me and at least two of us have tested it. Let's merge this baby and do some post-merge final testing to make sure everything works okay. We probably also want to add some docs about the feature to the site.

PR #74, opened 2013-05-29
Keep compass actions in compass task

robwierzbowski:
Compass generated images are handled by the Compass task and the copy task. This brings them all under the compass task.

Side note, generated images in the current gruntfile and after this PR aren't touched by imagemin.


sindresorhus:
@robwierzbowski thanks! sorry for the delay.

robwierzbowski:
No worries. I haven't tested this since I made the PR, so might want to double check it works with the current Gruntfile.

PR #75, opened 2013-05-29
Add touch icon pattern to copy:dist task

robwierzbowski:
Add root pngs to the copy:dist task to pick up default apple-touch-icons.


addyosmani:
Good call. Thanks!

PR #87, opened 2013-06-13
add jQuery globals to jshintrc

jeroentbt:
I had to manually add it to make jshint play nice with $.
Would be nice if it's included out of the box, I think.

### DIFF
--- a/app/templates/jshintrc
+++ b/app/templates/jshintrc
@@ -17,5 +17,6 @@
     "unused": true,
     "strict": true,
     "trailing": true,
-    "smarttabs": true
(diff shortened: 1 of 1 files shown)


passy:
jQuery is not part of the default scaffold, so I don't think we should add this option.

jeroentbt:
@passy. 
I checked that to be sure: (app/index.js, line 40)
"Out of the box I include HTML5 Boilerplate, jQuery and Modernizr."

passy:
Excuse me please, I'm confusing generators here. `webapp` does indeed include jQuery by default. :)

PR #90, opened 2013-06-14
Move require.js main.js to a template

### LINKED ISSUE #89: Freshly generated site fails linting due to extra comma in main.js
The generator creates a main.js containing a trailing comma in the last entry of the require.config.paths, e.g.

``` javascript
paths: {
    jquery: '../bower_components/jquery/jquery',
}
```

This is not valid in older versions of IE, and jshint will flag it as an error, thus failing the task. Not hard to modify main.js to fix, but I don't think generated sites should fail linting even before they have been modified?

passy:
This fixes #89


addyosmani:
This has been long overdue. Nice work, @passy!

PR #98, opened 2013-06-24
Use Usemin blocks by default.

stephenplusplus:
This is the same solution from https://github.com/yeoman/generator-angular/pull/248
- a Usemin block is created by default in the generated index.html file,
- the Concat task from the Gruntfile is commented by default, with a message explaining why.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -265,14 +265,20 @@ module.exports = function (grunt) {
             }
         },
         cssmin: {
-            dist: {
-                files: {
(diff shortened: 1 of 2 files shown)


addyosmani:
can we make '..will take care of' part of the same sentence and add an e.g at the end for the sample line? Other than that I think its fine.

stephenplusplus:
Updated! Let me know what you think.

addyosmani:
Perfect! :) looks good for a merge. Thanks for updating sir!

Issue #94, opened 2013-06-20
generator-webapp@0.2.4 : autoprefixer issue


fcamblor:
Seems like there is an issue with the `autoprefixer` task in `package.json` :

```
Loading "autoprefixer.js" tasks...ERROR
>> Error: Cannot find module 'autoprefixer'
```

I had to remove the following line from `package.json` to fix this :

```
    "grunt-autoprefixer": "~0.1.20130516",
```

(note that I answered `no` to the `Would you like to use autoprefixer for your CSS?` yeoman question)

passy:
As a quick fix, could you try and see if `grunt-autoprefixer` is listed in your `package.json` as `devDependency` and if so remove it?

This seems to be a problem with `autoprefixer` being a `peerDependency` of `grunt-autoprefixer`. It looks like sometimes npm fails to install this.

fcamblor:
Removing `grunt-autoprefixer` in package.json fixes the problem (as I stated in my initial post ;-))

passy:
D'oh, that comment was before my first coffee today. ;)

passy:
`grunt-autoprefixer` is no longer loaded through matchdep from now on, when it's disabled. However, I still don't understand why the peer dependency installation fails sometimes. Could you tell me what `npm --version` prints for you?

fcamblor:
Sorry for the bad latency :-)

```
$> npm --version
1.1.62
$> node --version
v0.8.11
```

addyosmani:
@passy with the change for grunt-autoprefixer no longer being loaded through matchdep, is this still occasionally failing? I ask as I believe this is one of the two issues we're waiting on before we can cut a new release.

passy:
@addyosmani We narrowed this down to obsolete npm versions. So this can only happen if you manually enable autoprefixer in the prompt and run a version of npm from before 1.2.10 or so. @sindresorhus suggested we should add an engine requirement to the generator for npm `~1.3.0` so that there is at least a warning if you're running an incompatible version.

addyosmani:
> @sindresorhus suggested we should add an engine requirement to the generator for npm ~1.3.0 so that there is at least a warning if you're running an incompatible version.

SGTM.

passy:
"Fixed" by #100

YonasBerhe:
I'm having issues when trying to run the grunt server command. 

yonas@yonas:~/Documents/code/vailroom/listingpage$ grunt serve
Running "serve" task

[...]

PR #100, opened 2013-06-25
Add engines requirement for npm >= 1.2.10

passy:
As suggested by @sindresorhus. This should at least give a hint to the users if they install the generator in an outdated environment. The warning they see looks like this:

screenshot from 2013-06-25 11 12 59

Related: #94


sindresorhus:
@stephenplusplus mentioned peerDependencies was introduced in 1.2.10, so maybe use that as a minimum instead? Can anyone confirm that?

addyosmani:
1.2.10 confirmed as when it was introduced as per http://blog.nodejs.org/2013/02/07/peer-dependencies/. I would favor this being used for the minimum too.

passy:
Sure. Give me a second.

addyosmani:
Thanks @passy!

sindresorhus:
Awesome.

This needs to be applied to all the generators and yo.

stephenplusplus:
`yo`
`generator-angular`
`generator-webapp`
`generator-backbone`
`generator-ember`
`generator-jquery`
`generator-generator`

^ have been hit with PRs for the update. Any other big ones we should target?

ragingwind:
@stephenplusplus :horse_racing:

PR #109, opened 2013-07-10
Fix duplicate CSS with autoprefixer

### LINKED ISSUE #108: CSS gets duplicated on build
This can be duplicated by doing the following in an empty folder:

`yo webapp` (i selected `no` to everything except autoprefixing)
`grunt`

View that the `main.css` file in the `dist` folder, while it's been minified, has been duplicated. 

If you throw in something that will be auto prefixed, the first 'copy' of the css file has been auto prefixed while the second has not been. I played around with the grunt file for a bit, but haven't had time to figure out the exact directive that is causing the problem. I'm assuming it's an order problem.

A quick fix is to rename the file in the app folder to main.scss and then it works just fine.

passy:
Fixes #108

> Due to the way the usemin block was set up, it would concat and minify both the
> non-prefixed and the prefixed CSS file. Now, depending on whether you enable
> autoprefixer or not, the usemin search path is adjusted and the snippet is
> programmatically inserted into the index.html template.

/cc @stephenplusplus @addyosmani Could you take a look that I did not introduce another silly mistake?

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -35,20 +35,19 @@ module.exports = function (grunt) {
             },
             compass: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
-                tasks: ['compass:server'<% if (autoprefixer) { %>, 'autoprefixer' <% } %>]
-            },<% if (autoprefixer) { %>
+                tasks: ['compass:server'<% if (autoprefixer) { %>, 'autoprefixer'<% } %>]
+            },
             styles: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.css'],
-                tasks: ['copy:styles', 'autoprefixer']
-            },<% } %>
+                tasks: ['copy:styles'<% if (autoprefixer) { %>, 'autoprefixer'<% } %>]
+            },
             livereload: {
                 options: {
                     livereload: LIVERELOAD_PORT
                 },
                 files: [
-                    '<%%= yeoman.app %>/*.html',<% if (autoprefixer) { %>
-                    '.tmp/styles/{,*/}*.css',<% } else { %>
-                    '{.tmp,<%%= yeoman.app %>}/styles/{,*/}*.css',<% } %>
+                    '<%%= yeoman.app %>/*.html',
+                    '.tmp/styles/{,*/}*.css',
                     '{.tmp,<%%= yeoman.app %>}/scripts/{,*/}*.js',
                     '<%%= yeoman.app %>/images/{,*/}*.{png,jpg,jpeg,gif,webp,svg}'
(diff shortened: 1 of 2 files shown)


stephenplusplus:
I used the default options during `yo webapp` and a blank `dist/styles/***.main.css` was generated for me :-\

``` html
<!-- build:css(app) styles/main.css -->
<link rel="stylesheet" href="styles/main.css">
<!-- endbuild -->
```

The `{app,.tmp}` was in there for Sass, since when useminPrepare does its thing, it finds `styles/main.css` in `.tmp/`, since `app/styles/main.css` doesn't exist. Which, I believe is what happened here.

passy:
Ugh, thanks for trying it out, Stephen. I'm glad I didn't just push it.

I missed the case when you would use Sass without autoprefixer. I guess the only way to make this work is to always have the copy task that moves the CSS over to `.tmp`.

passy:
I gave it another shot.

stephenplusplus:
I'm finally catching up to how `autoprefixer` works.
-webkit-I'm finally catching up to how `autoprefixer` works.
-moz-I'm finally catching up to how `autoprefixer` works.

That was lame, but I had to.

Everything works great for me now! :+1:

app/templates/Gruntfile.js
```diff
@@ -36,19 +36,18 @@ module.exports = function (grunt) {
             compass: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
                 tasks: ['compass:server'<% if (autoprefixer) { %>, 'autoprefixer' <% } %>]
-            },<% if (autoprefixer) { %>
+            },
             styles: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.css'],
-                tasks: ['copy:styles', 'autoprefixer']
-            },<% } %>
+                tasks: ['copy:styles'<% if (autoprefixer) { %>, 'autoprefixer' <% } %>]
```
stephenplusplus: Just being picky, but I think there's an extra space: `, 'autoprefixer'`<-- there.

passy: Whoa, thanks for catching that!

passy:
Haha, joke was good enough for me. :grin:

passy:
Thanks for helping out, @stephenplusplus! :cake:

addyosmani:
Thanks for fixing Passy!

On Wednesday, July 10, 2013, Pascal Hartig wrote:

> Thanks for helping out, @stephenplusplushttps://github.com/stephenplusplus!
> [image: :cake:]
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/yeoman/generator-webapp/pull/109#issuecomment-20775443
> .

## 

Sent from Gmail Mobile

PR #111, opened 2013-07-14
Make use of the new checkbox prompt

sindresorhus:
I think it looks better and is quicker to scan.

@addyosmani @passy @stephenplusplus what do you think?

screen shot 2013-07-14 at 18 25 36

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -39,30 +39,32 @@ AppGenerator.prototype.askFor = function askFor() {
   console.log('Out of the box I include HTML5 Boilerplate, jQuery and Modernizr.');
 
   var prompts = [{
-    type: 'confirm',
-    name: 'compassBootstrap',
-    message: 'Would you like to include Twitter Bootstrap for Sass?',
-    default: true
-  },
-  {
-    type: 'confirm',
-    name: 'includeRequireJS',
-    message: 'Would you like to include RequireJS (for AMD support)?',
-    default: true
-  },
-  {
-    type: 'confirm',
-    name: 'autoprefixer',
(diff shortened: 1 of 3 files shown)


sindresorhus:
Tests fails since we the mock prompt would need updating.

passy:
This is rad! :+1:

stephenplusplus:
This approach essentially mutes the "we can't have a prompt for everything" debate. 10 prompts with 1 option each isn't doable. 1 prompt with 10 options is.

:+1:

passy:
> This approach essentially mutes the "we can't have a prompt for everything" debate. 10 prompts with 1 option each isn't doable. 1 prompt with 10 options is.

Absolutely! Remembering the last choices would still be nice, but for the first run this is a much better approach.

addyosmani:
I think this is much easier to scan. Very nicely done :+1:  

Could you update the mock prompt too as a part of this PR? Happy to merge this soon after.

sindresorhus:
nvm, only the tests needed updating.

sindresorhus:
Oh, and since Autoprefixer has worked so good I've made it checked by default to get the last bit of feedback until we remove the prompt for it sometime in the future.

sindresorhus:
The other generators might benefit from something similar too. Hint hint ;)

Garbee:
Thanks for this @sindresorhus. It is a really nice update to handling Yeoman.

Very well done.

sindresorhus:
@Garbee I'm glad to hear :)

Do let us know if there are other things that could be improved.

PR #115, opened 2013-07-17
Drop Autoprefixer prompt, make it a default

addyosmani:
We haven't received many (if any) reports of issues with Autoprefixing since we introduced it. This PR drops the prompt for the feature and treats it as a default. 

Do let me know if I missed anything :)


passy:
:+1: You need to fix the tests, though. :)

addyosmani:
Closed by 803a385ced331844b92570756ac9397ca32bbb36

PR #116, opened 2013-07-21
Add `yo` as peerDependency

passy:
/re https://github.com/yeoman/generator/issues/305

Since `yo` specifies `grunt-cli` and `bower` as peerDependencies, this should allow a `npm install -g generator-webapp` to install all three of them globally. Is that right or am I missing something?

/cc @addyosmani @sindresorhus

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "cheerio": "~0.10.8"
   },
   "peerDependencies": {
-    "generator-mocha": "~0.1.1"
+    "generator-mocha": "~0.1.1",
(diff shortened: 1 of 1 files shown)


sindresorhus:
Ya, but do test just in case.

passy:
Verified against a clean npm dir. Installing `generator-webapp` also installs bower, yo and grunt.

sindresorhus:
<3 peerDependencies

sindresorhus:
Lets implement it on all the things

PR #123, opened 2013-08-08
add grunt-bower-install. closes #121.

### LINKED ISSUE #121: Add grunt-bower-install?
[grunt-bower-install](http://github.com/stephenplusplus/grunt-bower-install) seems to be performing well out there in the wild!

It's grown considerably, now doing deeper dependency matching and more accurate sorting. It also doesn't use `bower` internally anymore. `grunt bower-install:jquery` used to mask `bower install jquery --save`, but now, you just run `grunt bower-install` after installing/removing a Bower package, just like `grunt-bower-requirejs`.

Is it worth trying out `grunt-bower-install` if RequireJS isn't chosen by the user?

stephenplusplus:
This is an initial PR to add in `grunt-bower-install`, should we decide we want it in cases where a user opts to not use RequireJS.

Something unfortunate is I had to move Modernizr to a new `scripts/vendor.js` usemin block. The block in the generated `index.html` will look like this:

``` html
<!-- build:js scripts/vendor.js -->
<!-- bower:js -->
<script src="bower_components/modernizr/modernizr.js"></script>
<script src="bower_components/jquery/jquery.js"></script>
<!-- endbower -->
<!-- endbuild -->
```

Anything inside of the `<!-- bower:js -->` block will be replaced with the installed Bower packages. Since Modernizr is a Bower component, it will be shoved in. Without removing the former Modernizr include from `<head>`, it would be on the page twice.

From [Modernizr Docs](http://modernizr.com/docs/#installing):

> Drop the script tags in the `head` of your HTML. For best performance, you should have them follow after your stylesheet references. The reason we recommend placing Modernizr in the head is two-fold: the HTML5 Shiv (that enables HTML5 elements in IE) must execute before the `<body>`, and if you’re using any of the CSS classes that Modernizr adds, you’ll want to prevent a FOUC.
> 
> If you don't support IE8 and don't need to worry about FOUC, feel free to include modernizr.js whereever.

Let me know if you have any ideas for how we can implement this!

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -135,12 +135,26 @@ AppGenerator.prototype.writeIndex = function writeIndex() {
     '                <ul>'
   ];
 
+  var bowerInstallText
+    = '\n        <!-- build:js scripts/vendor.js -->'
+    + '\n        <!-- bower:js -->'
+    + (this.includeModernizr
+        ?
+      '\n        <script src="bower_components/modernizr/modernizr.js"></script>'
+        :
+      '')
+    + '\n        <script src="bower_components/jquery/jquery.js"></script>'
+    + '\n        <!-- endbower -->'
+    + '\n        <!-- endbuild -->'
+    + '\n';
+
   this.indexFile = this.readFileAsString(path.join(this.sourceRoot(), 'index.html'));
   this.indexFile = this.engine(this.indexFile, this);
 
   if (!this.includeRequireJS) {
+    this.indexFile = this.append(this.indexFile, 'body', bowerInstallText);
+
     this.indexFile = this.appendScripts(this.indexFile, 'scripts/main.js', [
-      'bower_components/jquery/jquery.js',
       'scripts/main.js'
     ]);
 
--- a/app/templates/_package.json
+++ b/app/templates/_package.json
@@ -15,8 +15,8 @@
     "grunt-contrib-clean": "~0.4.0",
     "grunt-contrib-htmlmin": "~0.1.3",<% if (includeRequireJS) { %>
     "grunt-bower-requirejs": "~0.4.1",
-    "grunt-contrib-requirejs": "~0.4.0",<% } %><% if (includeModernizr) { %>
-    "grunt-modernizr": "~0.3.0",<% } %>
+    "grunt-contrib-requirejs": "~0.4.0",<% } else { %>
+    "grunt-bower-install": "~0.5.0",<% } %>
     "grunt-contrib-imagemin": "~0.1.3",
     "grunt-contrib-watch": "~0.4.0",<% if (testFramework === 'jasmine') { %>
     "grunt-contrib-jasmine": "~0.4.2",<% } %>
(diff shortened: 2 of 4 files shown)


addyosmani:
Thanks for putting this together, @stephenplusplus! Let's keep this open for now. 

@sindresorhus @passy do we know what the current state of post-install hooks in the Bower API is? I'd love to be able to somehow call back to the generator system once `bower` has been run, ideally without the need to introduce grunt-bower-install (even though it is awesome).

If we don't think that is going to be possible for now, should `grunt-bower-install` be moved to a prompt or are we happy with it being considered as a default path?

passy:
/ping @satazor Any news regarding post-install hooks?

I think hooks would be nice, but this tool still provides great value, even when it has to be triggered manually.

Regarding Modernizr: Would it be possible to either provide a list of ignored packages in the Grunt configuration or  alternatively scan the complete HTML for script tags outside of `<!-- bower:js -->` tags and skip the inclusion of those files?

stephenplusplus:
Yep, either of those is brilliant.

satazor:
@passy yes there is a PR that implements it. Will need to review but it will be available in the next minor version

passy:
@satazor That's fantastic. Thanks for the update!

app/index.js
```diff
@@ -135,12 +135,26 @@ AppGenerator.prototype.writeIndex = function writeIndex() {
...
   ];
 
+  var bowerInstallText
+    = '\n        <!-- build:js scripts/vendor.js -->'
+    + '\n        <!-- bower:js -->'
+    + (this.includeModernizr
+        ?
+      '\n        <script src="bower_components/modernizr/modernizr.js"></script>'
+        :
+      '')
```
stephenplusplus, line 145: This probably couldn't look weirder. Thoughts? :)

app/templates/_package.json
```diff
@@ -24,7 +24,8 @@
     "grunt-autoprefixer": "~0.1.20130516",
     "grunt-usemin": "~0.1.10",<% if (testFramework === 'mocha') {  %>
     "grunt-mocha": "~0.3.0",<% } %>
-    "grunt-open": "~0.2.0",
+    "grunt-open": "~0.2.0",<% if (includeModernizr) { %>
+    "grunt-modernizr": "~0.3.0",<% } %>
```
stephenplusplus, line 28: Moved this block to avoid more confusion than there already is in this file :-S

app/templates/index.html
```diff
@@ -12,10 +12,7 @@
         <!-- Place favicon.ico and apple-touch-icon.png in the root directory -->
         <!-- build:css(.tmp) styles/main.css -->
         <link rel="stylesheet" href="styles/main.css">
-        <!-- endbuild --><% if (includeModernizr) { %>
-        <!-- build:js scripts/vendor/modernizr.js -->
-        <script src="bower_components/modernizr/modernizr.js"></script>
-        <!-- endbuild --><% } %>
+        <!-- endbuild -->
```
stephenplusplus, line 15: If the user wants Modernizr, it will create it in the `<body>` now. A future solution will be adding a feature to GBI to ignore specified packages, or not re-insert scripts that were already discovered on the page.

addyosmani, line 15: Granted we have a plan for improving this in the future, sounds okay for now :)

app/templates/_package.json
addyosmani, line 28: :+1:

app/index.js
addyosmani, line 145: Heh. Looks weird, but it works. I can't (ottomh) think of a better way to implement this. Btw, not necessarily for this PR, but we could replace the literal script tags with an array of ['modernizr/modernizr.js','jquery/jquery.js'] and just loop through them for their part of the bowerInstallText.

addyosmani:
Reviewed. Left a few comments for how we could improve this in future, but for now LGTM. Merging. Thanks @stephenplusplus! Have some :cake:

Issue #124, opened 2013-08-09
Discussion: make generator-webapp more mobile-friendly


passy:
I was talking to @gauntface today, who is using the webapp generator as part of his Cordova generator after the initial project structure is set up. This appears to work surprisingly well given that `generator-webapp` was never designed to be used as sub-generator to begin with.

However, he ran into a few issues that we could take care of:
- jQuery is a bad default for a mobile application. I think that it is less likely to be needed for a webapp today in general. It is part of H5BP, though, which we try to stay close to. Now that we have checkboxes, could we make it another optional setting?
- The default template isn't responsive. You have to scroll horizontally. That should be easy to fix.
- Installation and welcome messages should have an option to be turned off. When you call from another generator, you usually don't want to see them. Could be one or two additional flags, but it would make sense to keep those consistent across the default generators like `--skip-install`.

Thoughts?

/ping @addyosmani @sindresorhus @stephenplusplus @paulirish

addyosmani:
Good points!

So, @gauntface and I are exploring how to make generator-webapp more mobile-friendly as part of https://github.com/yeoman/generator-mobile and once we've done an alpha release I'd love to see if there's anything there that would benefit being brought into this generator.

I wouldn't mind moving jQuery to a prompt but we may want to survey developers before doing so. Most of the time when I need it these days it's to use with an accompanying framework so it makes more sense to continue bundling with those types of generators where it's a dependency. 

Regarding the installation messages, +1 on us having an option for those to be turned off. How about `--skip-welcome-message`, `--skip-install-message` or something along those lines?

gauntface:
Adding those messages sounds like the best approach to me.

I'll schedule this in for early next week and just put the relevant changes in as separate pull requests, should be fairly trivial changes.

gauntface:
Just to discuss the jQuery issue a little further, the best solution does seem to be to add further options to switch jQuery (and Modernizr) from being defaults to checkbox items.

This does slightly increase complexity, checking whether certain things are included or not now that they can be optional.

But it does cater for a use case where people may not think it's the best thing to include (and perhaps offer something else on top of webapp).

Would it be worth adding Y (include by default), N (No don't include by default), R (Remove from options list) as  parameters to a list of options (i.e. --jquery=R --modernizr=N --boilerplate=Y)

sindresorhus:
[...]

PR #125, opened 2013-08-10
Add --skip-install-message, --skip-welcome-message

passy:
Requires a currently unpublished version of `yeoman-generator > 0.13.0`.

/ref #124

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -23,7 +23,10 @@ var AppGenerator = module.exports = function Appgenerator(args, options, config)
   this.mainCoffeeFile = 'console.log "\'Allo from CoffeeScript!"';
 
   this.on('end', function () {
-    this.installDependencies({ skipInstall: options['skip-install'] });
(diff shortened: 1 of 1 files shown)


addyosmani:
LGTM. Can we publish the updates to the generator and then bring this in?

sindresorhus:
I don't like having more flags. Why do we need the skipInstallMessage one? And instead of the skipWelcome flag, can't we just move the ASCII to be shown on exit instead?

passy:
See #124. Flags are currently the best (only?) way to get integration with super-generators that want to call you. I do see the problem with those too that they create ugly nesting and need to be implemented for all generators separately.

sindresorhus:
Couldn't say no to Matts puppy eyes.

passy:
> Couldn't say no to Matts puppy eyes.

Don't blame yourself. Nobody can.

PR #128, opened 2013-08-11
Add prompt for grunt-modernizr

### LINKED ISSUE #107: Integrate grunt-modernizr
Should we integrate grunt-modernizr? Would be nice if we had a lean build based on what's actually used instead of shipping with the full-blown copy.

https://github.com/Modernizr/grunt-modernizr

passy:
Fix #107

This adds a prompt for `grunt-modernizr` and removes it completely, if the checkbox is unchecked (#124). I chose to directly overwrite the bower component for Modernizr, which seemed the easiest way. The grunt task is actually really well customizable, so if a detection doesn't work, you can still manually enable or disable certain checks or turn of the autodetection completely.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -329,7 +329,17 @@ module.exports = function (grunt) {
                 dest: '.tmp/styles/',
                 src: '{,*/}*.css'
             }
-        },
+        },<% if (includeModernizr) { %>
+        modernizr: {
+            devFile: '<%%= yeoman.app %>/bower_components/modernizr/modernizr.js',
+            outputFile: '<%%= yeoman.dist %>/bower_components/modernizr/modernizr.js',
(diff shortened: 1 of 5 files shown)


app/templates/Gruntfile.js
```diff
@@ -330,6 +330,16 @@ module.exports = function (grunt) {
                 src: '{,*/}*.css'
             }
         },
+        modernizr: {
```
sindresorhus, line 333: `<% if (includeModernizr) { %>`

passy, line 333: Woops, thanks.

sindresorhus:
WGTM

sindresorhus:
:horse:

passy:
:unicorn: (I'm sure it'll work one day.)

PR #131, opened 2013-08-14
put Modernizr back in <head>.

stephenplusplus:
RE: #123

I changed how [wiredep](https://github.com/stephenplusplus/wiredep), the magic behind grunt-bower-install, works based on @passy's recommendation (:smile_cat:). It will no longer insert a script that's already in the page, outside of the `<!-- bower:js -->..<!-- endbower -->` block. So, that means we can put Modernizr where it belongs, if the user wants it.

Sorry for the last minute change-up!


passy:
Wonderful, thank you!

addyosmani:
Danke!

PR #141, opened 2013-08-24
First attempt: Bootstrap 3 integration

### LINKED ISSUE #134: Upgrade to Bootstrap 3
http://blog.getbootstrap.com/2013/08/19/bootstrap-3-released/
- [x] Bump revision
- [x] Update references to plugins
- [x] Update markup
- [x] Make sure Glyphicons work
- [ ] Eat popsicles :dango:

passy:
As @schickling told me, `sass-bootstrap` does indeed already have a 3.0, so I played around with it a little and came up with this:

download
- Name in the top right corner is the `appName`
- Not sure about the menu. Do we want to keep something like this in? Most apps probably want to have one, right?
- The glyphicons is just for fun and isn't generated.
- Features are below.
- Generated CSS is [from the example](http://getbootstrap.com/examples/jumbotron-narrow/).

(fixes #134)


schickling:
Nice! Thanks for your work!

But isn't here the same problem as described in https://github.com/yeoman/generator-angular/pull/339#issuecomment-23217200 ?

And another question: Is the icon png really needed? I thought BS3 now uses icon fonts?

passy:
@schickling It's a little trickier in `generator-angular`, because it supports Bootstrap without Compass, while `generator-webapp` doesn't have an option to work with vanilla CSS. And what do you mean with the icon png? I thought I had removed all references to it.

schickling:
1. Bootstrap: Alright. Is it intend that `generator-angular` uses Bootstrap without Compass and `generator-webapp` plain CSS or will that also be adjusted?
2. Icons: My bad :smile:

passy:
@schickling It's probably going to stay that way for the near future, unless someone finds a really sexy way to bring this into webapp as well. But it probably makes sense if we figure out how to upgrade to 3.0 in one generator at a time, so we don't duplicate the work unnecessarily.

passy:
Updated for 6b029ecbaa72d8eb11117f55a0cea48874eabccf.

@sindresorhus @addyosmani Review me maybe?

kevva:
lgtm. Much nicer to keep the HTML and JS separate too.

sindresorhus:
Oh yeah, this is dexterous!

Tried running `grunt`, but Phantom seems to be timing out. Anyone else getting that?

```
Running "mocha:all" (mocha) task
Testing: http://localhost:9000/index.html

Warning: PhantomJS timed out, possibly due to a missing Mocha run() call. Use --force to continue.

Aborted due to warnings.
```

Can you fix the merge conflict?

app/templates/index.html
```diff
@@ -18,10 +18,68 @@
...
+                    <% } %>
+
+                    <% if (includeRequireJS) { %><h4>RequireJS</h4>
+                    <p>RequireJS is a JavaScript file and module loader. It is optimized for in-browser use, but it can be used in other JavaScript environments, like Rhino and Node.</p>
+                    <% } %>
+                </div>
+            </div>
+
+            <div class="footer">
+                <p>&lt;3 from the Yeoman Team</p>
```
sindresorhus, line 61: unicode: ♥

and shouldn't the `t` be lowercase?

addyosmani:
This is really nice work. Well done @passy!

In a perfect world, this is something we could put to a `generator-bootstrap` that was called by `generator-webapp` or any other generator that wished to do something with Bootstrap-based scaffolding. It would be quicker for just to just land this in its current form once the comments have been addressed and merge conflicts resolved though.

passy:
@sindresorhus Hm, works for me. PhantomJS installation broken?

@addyosmani Removing is definitely easier than adding (which is a pain even on the generator level), but I'm definitely :+1: on adding some way for the users to better pick and chose what they want. And having a sub-generator that would handle all the bootstrap stuff would be amazing, especially since AngularJS wants the exact same thing, but we have to do the work twice.

addyosmani:
Yay!

addyosmani:
I think a bootstrap sub-generator would make a great deal of sense. It could be general enough to support configuring whether or not to use just CSS or sass-bootstrap and could contain placeholders for other generators to fill in with content. 

e.g `generator-webapp` would populate with the list of what has installed, another generator may wish to fill it in with its own installed items or something else.

How would we see something like this being consumed by other generators?

schickling:
Excellent idea @addyosmani :+1:

PR #142, opened 2013-08-25
require_main.js: ".js" should not be included in config script paths

robwalch:
".js" should not be included in script paths.

Currently, if you try to require one of the bootstrap scripts `require(['app', 'jquery', 'bootstrapTooltip']` an error is thrown

when you try to use it

```
$().tooltip();

// Uncaught TypeError: Object [object Object] has no method 'tooltip'
```

or by requirejs

```
Request URL:http://mewmew.local:9000/bower_components/sass-bootstrap/js/bootstrap-tooltip.js.js

Uncaught Error: Script error for: bootstrapTooltip
http://requirejs.org/docs/errors.html#scripterror 
```

### DIFF
--- a/app/templates/require_main.js
+++ b/app/templates/require_main.js
@@ -1,17 +1,17 @@
 require.config({
     paths: {
         jquery: '../bower_components/jquery/jquery'<% if (compassBootstrap) { %>,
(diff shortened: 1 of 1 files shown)


sindresorhus:
You're right. Thanks :)

PR #159, opened 2013-09-10
Added Bootstrap modal & internal deps to require.js

HaNdTriX:
bootstrapTransition is not required in order make most js components to work, but most of the times people forget to include them. Later they wonder, why the transitions do not work.

[bootstrapPopover](https://github.com/twbs/bootstrap/blob/master/js/popover.js#L30) does not work without bootstrapTooltip.


HaNdTriX:
thanks!

PR #165, opened 2013-09-19
Prevent compass adding cachebuster to assets

callumlocke:
Compass adds a query-string cachebuster to CSS assets, eg `/images/foo.png?12342134123`.

I think we should disable this in webapp, because:
1. it's redundant – we already use revving for cachebusting (in the actual filename)
2. it's not necessarily harmless – the mere presence of a query string can cause some CDNs and proxies to cache less aggressively


sindresorhus:
Good point

addyosmani:
:+1:

passy:
:+1:

PR #180, opened 2013-10-16
added .sass-cache to grunt clean:*

tomfuertes:


sindresorhus:
nope, we shouldn't clean that.

PR #181, opened 2013-10-18
Make sure Grunt knows to compile Literate CoffeeScript too

cobyism:
No idea if this is a legit problem or not, or whether it’s just me missing something obvious, but the default Gruntfile provided by this generator didn’t seem to be picking up any files with `.litcoffee` or `.coffee.md` extensions. It looked to me like the Gruntfile template itself needed updating in order to support this, so this pull requests adds `.litcoffee` and `.coffee.md` extensions to the search paths for CoffeeScript files. Please let me know if there’s a better way to make this work, or if there’s another place I should be submitting this instead :grin:.

Also, the tests for this package seem to only really care about which files the generator places when you run it, so I wasn’t sure if it’d be in-scope to test whether the generated Gruntfile actually compiles the desired extensions or not. Let me know if that’s something that I should try and add in here too (my knowledge of how to test stuff like that in node is pretty much non-existent though… :sweat_smile:).

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -21,11 +21,11 @@ module.exports = function (grunt) {
         },
         watch: {<% if (coffee) { %>
             coffee: {
-                files: ['<%%= yeoman.app %>/scripts/{,*/}*.coffee'],
+                files: ['<%%= yeoman.app %>/scripts/{,*/}*.{coffee,litcoffee,coffee.md}'],
                 tasks: ['coffee:dist']
             },
             coffeeTest: {
-                files: ['test/spec/{,*/}*.coffee'],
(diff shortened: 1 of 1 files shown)


cobyism:
@addyosmani As described in [this blog post](http://ashkenas.com/literate-coffeescript/) by @jashkenas:

> Give this file a `.litcoffee` or a `.coffee.md` extension, and the CoffeeScript compiler will happily compile it to JavaScript for you — or run it directly.

... both extensions are valid indications that a file is literate coffeescript. You can also see the regex used for this in [the coffeescript source itself](https://github.com/jashkenas/coffee-script/blob/master/src/helpers.coffee#L131-135) too.

sindresorhus:
Thanks :)

cobyism:
:metal: :sparkles:

PR #193, opened 2013-11-06
Use grunt-newer

### LINKED ISSUE #151: use grunt-newer
This an essential missing piece in grunt. Luckily for us @tschaub created a task to only run tasks on changed files.

https://github.com/tschaub/grunt-newer

I did a quick test and looks like it would bring our build time from 10s to 2s. Though a much more important place for it is with `grunt server`.

tschaub:
This brings in the `grunt-newer` task.

There's not really room to speed things up in the `build` task because `clean:dist` is run first (so we always have to process all files again).

Open to other suggestions where this might be put to use.

Closes #151.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -34,7 +34,7 @@ module.exports = function (grunt) {
             },
             styles: {
                 files: ['<%%= yeoman.app %>/styles/{,*/}*.css'],
-                tasks: ['copy:styles', 'autoprefixer']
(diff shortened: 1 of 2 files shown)


eddiemonge:
this seems like it would be good for `grunt serve` according to a PR in generator-angular

addyosmani:
I think there's room for a more depth analysis of where we can take advantage of this. @passy do you have insights into how practically useful this is in the PR on Angular?

passy:
@addyosmani Applied it to `generator-angular` :)

Issue #191, opened 2013-11-06
--test-framework <framework> should be --test-framework=<framework>


jdelight:
From the readme, missing an equals sign  :)

sindresorhus:
Equal sign is optional.

jdelight:
Ahh ok thanks - didn't seem to work for me but I'll try again. Also having recently used the ember generator where the docs have a very similar but slightly different command it would be nice to use the same format to reduce confusion.
https://github.com/yeoman/generator-ember

readme for generator-ember:
--test-framework=[framework]

readme for generator-webapp:
--test-framework &lt;framework&gt;

minor point I know - but consistency would be nice.

Also really appreciate all your work for the web community! So thanks :)

paulRbr:
@sindresorhus I indeed had a problem specifying the test-framework from command line, see https://github.com/popox/generator-webapp-rjs/issues/2

and

my PR

PR #196, opened 2013-11-13
Need test-framework generators as dependencies for generated app

### LINKED ISSUE #120: Adding test-framework attribute without equals sign returns error
When executing `yo webapp --test-framework jasmine` the generator works until it tries to install jasmine. I receive the following from the console:

```
Error webapp --test-framework jasmine 

You don't seem to have a generator with the name true:app installed.
```

However, everything works fine if an equals sign is used i.e. `yo webapp --test-framework=jasmine`. I think the README needs to reflect this.

OSX 10.8.3

paulRbr:
There is a need to include the test-framework dependencies generators in the generated `package.json`

This fixes the issue #191 which was a real issue :)

EDIT: and fixes #120

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -8,14 +8,13 @@ var yeoman = require('yeoman-generator');
 var AppGenerator = module.exports = function Appgenerator(args, options, config) {
   yeoman.generators.Base.apply(this, arguments);
 
+   // for hooks to resolve on mocha by default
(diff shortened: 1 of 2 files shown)


eddiemonge:
The framework generators should not be included in the package. If people want to use those generators they should install them separately and preferably globally.

The other fix I don't feel like is the correct fix for the problem

PR #201, opened 2013-11-17
reload tests when test files are saved

### LINKED ISSUE #198: bootstrap-sass-official version 3> path error
version gulp-webapp 0.1.0
To reproduce after creating app that includes bootstrap
update bower to use the latest version of bootstrap-sass-official which is currently version 3.2.0
run gulp and you get a path error
The latest version of bootstrap-sass-official from bower has removed the vendor directory, looks like this is true for all bootstrap versions 3>
dir structure used to be
--bootstrap-sass-official
  --vendor
    --assets
but with the 3.2.0 version it is
--bootstrap-sass-official
  --assets

eddiemonge:
fixes https://github.com/yeoman/generator-webapp/issues/198

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -26,7 +26,11 @@ module.exports = function (grunt) {
             },
             coffeeTest: {
                 files: ['test/spec/{,*/}*.{coffee,litcoffee,coffee.md}'],
-                tasks: ['coffee:test']
+                tasks: ['coffee:test', 'test:watch']
+            },<% } else { %>
+            jstest: {
+                files: ['test/spec/{,*/}*.js'],
+                tasks: ['test:watch']
             },<% } %>
             compass: {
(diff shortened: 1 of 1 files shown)


addyosmani:
LGTM. Passy is going to do a second take.

passy:
@eddiemonge How should this be used? Is that something we should add to the README? If `grunt test watch` is the supposed way, I think we need to move the `connect` part to the not-watch condition. Otherwise it will fail because the connect server is already running.

eddiemonge:
it runs as part of `grunt serve`. When you save a spec file, it will run the unit tests. The connect:test can run alongside the normal connect as it runs on a different port. It starts up, runs the tests and then quits.

If you do `grunt test` it will run everything normally.

passy:
Gotcha. I think I was too impatient to see the result. Working nicely!

PR #203, opened 2013-11-17
add comments to gruntfile

eddiemonge:


app/templates/Gruntfile.js
```diff
@@ -306,7 +341,10 @@ module.exports = function (grunt) {
                 dest: '.tmp/styles/',
                 src: '{,*/}*.css'
             }
-        },<% if (includeModernizr) { %>
+        },
+
+        <% if (includeModernizr) { %>
+        // Smaller Modernizr builds optimized for your site
```
passy: Maybe: "Create a custom Modernizr build including just the checks you use in your app"?

passy: *Creates

passy:
I like this. @addyosmani mentioned a couple of times that even if you don't use the webapp generator, it still makes a great reference for grunt tasks. Having a good comment coverage certainly helps with that.

app/templates/Gruntfile.js
eddiemonge: Creates a custom Modernizr build of just the checks you use in your app
They all sound odd to me. @kswedberg suggestions?

kswedberg: @eddiemonge: Here's a tiny tweak of yours:

Creates a custom Modernizr build that includes only the tests you reference in your app.

Or maybe "Generates" instead of "Creates"?

sindresorhus:
Very nice @eddiemonge :)

We should have done this from the start.

addyosmani:
Really appreciate the work you put into this. So nice to have it in place. We should do a tweet to let folks know improved comments are in there. At least mention in the next newsletter update.

oskarrough:
Yes, thank you very much!

PR #214, opened 2013-12-06
Fix #190: modernizr task

### LINKED ISSUE #190: Modernizr task doesn't seem to serve a purpose.
The grunt modernizr task which triggers during a build doesn't seem to be plumbed in properly. It produces a compact custom modernizr script in a `bower_components` folder within `dist`, which isn't linked to any of the project files and doesn't replace the existing `vendor.js` concatenated file in the header.

Is there a way to automate concatenating this back into the vendor.js file and adding the cache busting seed number to to the end of the file? e.g. vendor.3532523.js

https://github.com/yeoman/generator-webapp/blob/master/app/templates/Gruntfile.js#L307-L316

passy:
The Modernizr inclusion should happen outside of a usemin block, since the task
itself takes care of uglification and should not be replaced by the full build.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -434,9 +434,9 @@ module.exports = function (grunt) {
         'autoprefixer',
         'concat',
         'cssmin',
-        'uglify',<% if (includeModernizr) { %>
+        'uglify',
+        'copy:dist',<% if (includeModernizr) { %>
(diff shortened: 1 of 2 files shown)


eddiemonge:
Looks good and works. Not sure about speed improvements since time-grunt is displaying for any task for me

PR #217, opened 2013-12-09
Use `grunt-bower-install` for CSS dependencies.

stephenplusplus:
@gojohnnygo made a good point in https://github.com/stephenplusplus/grunt-bower-install/issues/25#issuecomment-30177128 ... the webapp generator ships with JS support for `grunt-bower-install`, but gbi now also handles CSS deps as well.

Thought I'd get a few opinions on this before brute merging it in.

One thought I had was only using gbi for CSS if the user didn't opt into compass. Would that make sense, or not really?

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -10,6 +10,10 @@
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width">
         <!-- Place favicon.ico and apple-touch-icon.png in the root directory -->
+        <!-- build:css styles/vendor.css -->
+        <!-- bower:css -->
+        <!-- endbower -->
+        <!-- endbuild -->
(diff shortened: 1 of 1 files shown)


addyosmani:
@stephenplusplus 

I :love_letter: it. Imo, I wouldn't couple the decision to use gbi here to whether or not the user opted into using Compass, but that's just my thought on the matter. I'd like to go ahead and merge this in. Testing doesn't show any breakage during use.

Side: If you have bandwidth over the next few weeks/later in the new year, would love to also chat about gbi support for Polymer/Web Components. Currently doing something uber manual over on generator-polymer that would benefit from the task.

stephenplusplus:
Sounds awesome!

PR #218, opened 2013-12-10
Remove version, add private to bower.json template

alefteris:
Set the private property in bower.json template and remove the optional version property, as you wouldn't normally want to publish a webapp to the bower package registry.


addyosmani:
LGTM. Confirmed with Simon and Sindre that we want this. Thanks!

kvnwolf:
Why is this useful?

PR #234, opened 2013-12-17
Main.js file location and script order location

### LINKED ISSUE #233: main.js empty when using coffescript
I was using the generator with coffeescript.
However, when doing a build, the resulting main.js in the dist folder is empty.
Using serve works, using serve:dist does not work.

eddiemonge:
Fixes #233


sindresorhus:
Could you add a test?

eddiemonge:
Updated the title to reflect what the pr actually does.

only tests I can think of for this are to check indexOf for position of scripts in the index file. would that be enough or did you have some other tests in mind?

sindresorhus:
Nah, that would be too fragile, nvm :)

Issue #237, opened 2014-01-05
Refactoring CSS Pre-Processing


haydenbleasel:
I've been thinking about the way Yeoman handles stylesheets lately, especially with regards to preprocessing with languages / frameworks like Less, Sass and Compass. I have some thoughts:
#### 1) Make the Compass task optional

I'm not going to discuss this point in detail since there's already a popular issue on [making the Compass task optional](https://github.com/yeoman/generator-webapp/issues/114), but the general proposal is to remove the dependancy on Compass with regards to the second issue...
#### 2) Remove Bootstrap from the Sass option

The current Yeoman scaffold includes an option for Bootstrap for Sass. Selecting this option informs the scaffold uses Sass / Compass as the default stylesheet language (main.css -> main.scss).

My proposal is to remove Bootstrap from the Sass option, so the option is simply "Use Sass pre-processing". If the user wants Bootstrap for Sass, they can get it through Bower on their own accord. This also helps since the user may want to use Sass but not necessarily implement Bootstrap.
#### 3) Include Less as a pre-processor option

The title should be fairly clear, but adding Less as a CSS pre-processing option would be a good idea. I guess options for this are [grunt-contrib-less](https://github.com/gruntjs/grunt-contrib-less) or [grunt-recess](https://github.com/sindresorhus/grunt-recess).
#### TL;DR

Options for using Less and Sass without the need for Bootstrap included and removing the dependancy on Compass.

addyosmani:
I agree that we can improve the current CSS preprocessor setup, however, I personally disagree about removing Bootstrap outright however. When you're starting a new project, especially during prototyping, having a set of default styles out of the box can be pretty powerful. Perhaps break Sass/Compass/Bootstrap out into prompts that make sense rather than dropping features?

sindresorhus:
What Addy said. This generator is meant to be an opinionated starting point for creating web apps.

Maybe prompts like:
- Sass
- Bootstrap
- Modernizr

## 

Re LESS, see https://github.com/yeoman/yeoman/issues/289 We prefer Sass, but it should be easy to switch that out with LESS.

haydenbleasel:
[...]

Issue #114, opened 2013-07-16
Make the Compass grunt task optional


asapach:
`compass` task assumes the presence of Compass and Ruby in the path, which is not an explicit prerequisite. The task runs even when the project doesn't contain any SASS files and fails as the result:

> Warning: Running "compass:dist" (compass) task
> Warning: You need to have Ruby and Compass installed and in your system PATH for this task to work. More info: https://github.com/gruntjs/grunt-contrib-compass Use --force to continue.

I believe Compass should be an optional component in the template.

sindresorhus:
Yes, that's planned.

addyosmani:
Both Compass and CoffeeScript are examples of features we're considering making optional. The piece of work we are waiting on to land before they move to prompts is https://github.com/yeoman/generator/issues/277.

passy:
I think we should consider removing the prompt for autoprefixer before making compass optional to reduce the complexity of the Gruntfile. The CSS pipeline is already super confusing.

sindresorhus:
I would like for us to do this soon.

I think we should also consider bringing in grunt-sass as an alternative to Compass (though with visible warning that some Sass features might not be implemented). grunt-sass is way faster than compass, and would really help speed up livereload to be instantaneous instead of taking 2s.

addyosmani:
@sindresorhus do you have a link to a thread or somewhere where perf benchmarking was done between the two? I'm really interested in seeing whether the speed improvements are really that drastic. It would be pretty awesome.

robwierzbowski:
Perf differences are huge between Ruby Sass and libsass. Around 3 sec to 0.3 sec IIRC. But libsass doesn't contain some basic language features (like @extends), and can't be extended with Compass or Compass plugins (which expose new Ruby based functions in Sass stylesheets). Most Sass users take features over speed in my experience, or switch to Stylus.

Nothing against adding libsass, just providing some background.

sindresorhus:
What @robwierzbowski said.

I tried to get some comparison on it and more, but didn't get nay traction: https://github.com/h5bp/lazyweb-requests/issues/117

eddiemonge:
This has been fixed in #238

PR #238, opened 2014-01-06
Refactoring CSS Pre-Processing

haydenbleasel:
Pull request for [this issue](https://github.com/yeoman/generator-webapp/issues/237) consists of 2 major changes:

1) Binding Compass functions to the compassBootstrap (now includeCompass) variable. As mentioned in my issue, there was another relatively popular issue on [making the Compass grunt task optional](https://github.com/yeoman/generator-webapp/issues/114). Compass functions are now only included if the user selects to use Sass pre-processing.

2) Separate Bootstrap from Sass with Compass. This was a major change discussed between @addyosmani, @sindresorhus and myself. Rather than the previous `Bootstrap with Sass` option, the prompt is now:

```
[] Sass with Compass
[] Bootstrap
[] Modernizr
```

which results in the following 4 workflows:
- Just CSS
- Just Sass
- Bootstrap with CSS
- Bootstrap with Sass

Each workflow has been tested with `grunt` and `grunt serve` functions.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -48,8 +48,12 @@ AppGenerator.prototype.askFor = function askFor() {
     name: 'features',
     message: 'What more would you like?',
     choices: [{
-      name: 'Bootstrap for Sass',
-      value: 'compassBootstrap',
+      name: 'Sass with Compass',
+      value: 'includeCompass',
+      checked: true
+    }, {
+      name: 'Bootstrap',
+      value: 'includeBootstrap',
       checked: true
     }, {
       name: 'Modernizr',
@@ -65,7 +69,8 @@ AppGenerator.prototype.askFor = function askFor() {
 
     // manually deal with the response, get back and store the results.
     // we change a bit this way of doing to automatically do this in the self.prompt() method.
-    this.compassBootstrap = hasFeature('compassBootstrap');
+    this.includeCompass = hasFeature('includeCompass');
+    this.includeBootstrap = hasFeature('includeBootstrap');
     this.includeModernizr = hasFeature('includeModernizr');
 
     cb();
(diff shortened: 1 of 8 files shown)


addyosmani:
Thanks for putting this together. 

In the short term, I favor this set of changes for improving the flexibility of our Sass/Bootstrap setup.

It does introduce one more prompt, but I wouldn't mind it given how much we've shaved off the prompts queue in recent months.

addyosmani:
Landed \o/ 

@sindresorhus @passy I think I've mentioned before, but I still have some reservations about us auto-including in the Bootstrap plugins when you select Bootstrap. Unsure if others feel the same pain, but I always end up removing them manually.

passy:
Woo! \o/

Maybe @stephenplusplus has an idea if we can utilize `grunt-bower-install` somehow to ease the plugin pain a bit.

haydenbleasel:
@addyosmani - Thanks for accepting the PR. We should be able to close [this issue](https://github.com/yeoman/generator-webapp/issues/114) now as well. My first commit was binding Compass to the Sass with Compass option in the prompt.

I'll see if I can help with some of the pressing issues that are open and look for more ways to improve Yeoman. What's the status on documentation these days?

sindresorhus:
@addyosmani yeah, i never use them either and was previously opposed to having them in by default.

addyosmani:
@haydenbleasel We could use all the help with documentation you could offer. I'd be happy to jump into a new issue or Google doc to talk about some of the ways this could be improved.

Are you looking for smaller tasks or do you think you might have time over the next few weeks for more? :)

addyosmani:
@sindresorhus @passy I can open up an issue for us to either a) drop the plugins or b) move them to their own prompt. Another option is for us to ask on twitter if anyone uses the plugins while scaffolding.

sindresorhus:
@addyosmani do it. this would be so much easier if we used Insight for prompt choices ;)

haydenbleasel:
@addyosmani @sindresorhus - Moved this convo to a new issue.

Also Addy, new issue for improving documentation? There's still [this issue](https://github.com/yeoman/yeoman/issues/829)

PR #245, opened 2014-01-20
separates task configs to files

creynders:
Grunt files tend to grow huge and illegible **very** fast, it would be a real improvement if the task configurations would be separated to files. I chose for a `grunt` directory to add all config files to, but maybe that's not the best choice.

The loadConfig function was copied from https://gist.github.com/thomasboyt/6406507#file-load_config-js, I added a comment as attribution, but I don't know if that's good practice.

Hope you guys accept this one, because the monolithic grunt files drive me crazy :wink:

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -79,6 +79,35 @@ AppGenerator.prototype.askFor = function askFor() {
 
 AppGenerator.prototype.gruntfile = function gruntfile() {
   this.template('Gruntfile.js');
+  this.template('grunt/autoprefixer.js');
+  this.template('grunt/bower-install.js');
+  this.template('grunt/clean.js');
+  this.template('grunt/concurrent.js');
+  this.template('grunt/connect.js');
+  this.template('grunt/copy.js');
+  this.template('grunt/htmlmin.js');
+  this.template('grunt/imagemin.js');
+  this.template('grunt/jshint.js');
+  this.template('grunt/rev.js');
+  this.template('grunt/svgmin.js');
+  this.template('grunt/usemin.js');
+  this.template('grunt/useminPrepare.js');
+  this.template('grunt/watch.js');
+  
+  if (this.testFramework === 'mocha'){
+    this.template('grunt/mocha.js');
+  }else if(this.testFramework === 'jasmine'){
+    this.template('grunt/jasmine.js');
+  }
+  if(this.coffee){
+    this.template('grunt/coffee.js');
+  }
+  if(this.includeCompass){
+    this.template('grunt/compass.js');
+  }
+  if(this.includeModernizr){
+    this.template('grunt/modernizr.js');
+  }
 };
 
 AppGenerator.prototype.packageJSON = function packageJSON() {
--- a/app/index.js
+++ b/app/index.js
@@ -30,6 +30,7 @@ var AppGenerator = module.exports = function Appgenerator(args, options, config)
   this.options = options;
 
   this.pkg = JSON.parse(this.readFileAsString(path.join(__dirname, '../package.json')));
+  this.gruntpath = JSON.parse(this.readFileAsString(path.join(this.sourceRoot(), 'gruntrc'))).directory;
 };
 
 util.inherits(AppGenerator, yeoman.generators.Base);
@@ -78,35 +79,36 @@ AppGenerator.prototype.askFor = function askFor() {
 };
 
 AppGenerator.prototype.gruntfile = function gruntfile() {
+  this.copy('gruntrc', '.gruntrc');
   this.template('Gruntfile.js');
-  this.template('grunt/autoprefixer.js');
-  this.template('grunt/bower-install.js');
-  this.template('grunt/clean.js');
-  this.template('grunt/concurrent.js');
-  this.template('grunt/connect.js');
-  this.template('grunt/copy.js');
-  this.template('grunt/htmlmin.js');
-  this.template('grunt/imagemin.js');
-  this.template('grunt/jshint.js');
-  this.template('grunt/rev.js');
-  this.template('grunt/svgmin.js');
-  this.template('grunt/usemin.js');
-  this.template('grunt/useminPrepare.js');
-  this.template('grunt/watch.js');
-  
+  this.template(this.gruntpath+'/autoprefixer.js');
+  this.template(this.gruntpath+'/bower-install.js');
+  this.template(this.gruntpath+'/clean.js');
+  this.template(this.gruntpath+'/concurrent.js');
+  this.template(this.gruntpath+'/connect.js');
+  this.template(this.gruntpath+'/copy.js');
+  this.template(this.gruntpath+'/htmlmin.js');
+  this.template(this.gruntpath+'/imagemin.js');
+  this.template(this.gruntpath+'/jshint.js');
+  this.template(this.gruntpath+'/rev.js');
+  this.template(this.gruntpath+'/svgmin.js');
+  this.template(this.gruntpath+'/usemin.js');
+  this.template(this.gruntpath+'/useminPrepare.js');
+  this.template(this.gruntpath+'/watch.js');
+
   if (this.testFramework === 'mocha'){
-    this.template('grunt/mocha.js');
+    this.template(this.gruntpath+'/mocha.js');
   }else if(this.testFramework === 'jasmine'){
-    this.template('grunt/jasmine.js');
+    this.template(this.gruntpath+'/jasmine.js');
   }
   if(this.coffee){
-    this.template('grunt/coffee.js');
+    this.template(this.gruntpath+'/coffee.js');
   }
   if(this.includeCompass){
-    this.template('grunt/compass.js');
+    this.template(this.gruntpath+'/compass.js');
   }
   if(this.includeModernizr){
-    this.template('grunt/modernizr.js');
+    this.template(this.gruntpath+'/modernizr.js');
   }
 };
 
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -39,8 +39,10 @@ module.exports = function (grunt) {
         }
     }
 
+	var configsPath = grunt.file.readJSON('.gruntrc').directory;
+
     // Define the configuration for all the tasks
-    grunt.util._.extend(config, loadConfig('./grunt/'));
+    grunt.util._.extend(config, loadConfig(configsPath));
 
     grunt.initConfig(config);
 
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -7,6 +7,21 @@
 // use this if you want to recursively match all subfolders:
 // 'test/spec/**/*.js'
 
+
+//based on https://gist.github.com/thomasboyt/6406507#file-load_config-js
+function loadConfig(path){
+    var glob = require('glob');
+    var object = {};
+    var key;
+
+    glob.sync('*', {cwd : path}).forEach(function(option){
+        key = option.replace(/\.js$/, '');
+        object[key] = require(path + option);
+    });
+
+    return object;
+}
+
 module.exports = function (grunt) {
 
     // Load grunt tasks automatically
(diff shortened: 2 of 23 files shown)


creynders:
Personally I'd opt for `config`, `options` or `grunt`. I don't quite get the *modules or `dependencies` naming, sounds misleading to me? 
And yes, I didn't know what to do with the spacing. The `.editorconfig` declares 2 spaces, yet the gruntfile uses 4, so I went with 4 too.

creynders:
@thomasboyt your gists didn't contain any license info, do you have a preference on attribution?

addyosmani:
I and the team have experimented with split gruntfiles in the past (I use this pattern in at least one personal generator). Imo, it becomes a challenge modifying and maintaining shared config across files once your project grows.

That said, we'll keep this open so the community have an opportunity to give us their feedback and comments. Maybe we're wrong :)

addyosmani:
Also, major kudos for the PR!

creynders:
@addyosmani how exactly does it become cumbersome? I had some trouble the first time I used split configurations, but they were all due to potentially confusing/conflicting env and config property names and not having been consistent in using a config object for all configurable paths, for instance. 

Speaking of which, there are some optimizations that could be done in that regard. Calling a config object `yeoman` doesn't make sense IMHO, since it doesn't communicate what data it holds.
I tend to have a single `config` object with properties like `pkg`, `env` and `paths` (which in turn store all configurable contextual data). Haven't had any maintainability problems ever since.

If you want I can create a PR as a proposition.

sindresorhus:
I agree with @addyosmani. I've tried it out and found it to be even more cumbersome than a large Gruntfile. I think rather than splitting up the config, we should try to reduce it. I like your suggestion for better config property name though. The `yeoman` comes from legacy decisions and should be changed to something more obvious like you've suggested.

creynders:
Could you elaborate on that @sindresorhus? As I said, I've been using split configurations for a while now and IMO it's very easy to use, but obviously my situation and approach would be different of yours, so I'm curious in what way it complicates your workflow.

sindresorhus:
Our config is very coupled and splitting it up makes it even harder to comprehend how it all relates and works.

When I open the Gruntfile I can see everything and quickly move between different parts of the config. Not so much when all the config is split over various files. I just don't see any benefit.

eddiemonge:
if this were to happen, I'd rather see the config files as yml, or at least json to make it easier to read/edit/comprehend.

creynders:
@sindresorhus I understand that, but isn't that a problem which arose since the configs weren't split up to begin with? I.e. it's hard to split up and maintain the configs if you started out with the monolithic grunt file?  

The benefits of split configurations are: 
1. it's a lot faster to modify a task config when it's in a separate file, no searching through a 1000 LOC file 
2. Task configuration modification commits "read" a lot easier. Viewing diffs in a monolithic grunt file doesn't show what task config has changed if the modification is many lines below the task name.
3. Task configurations can be easily reused and synced. For instance, the project I'm working on I have a `dist` which in turn has a Grunt file of it's own (which is a lot smaller than the "main" one and has specific custom configurations, but also reuses some of the main one) At first it was a pain to keep some of the task configs synced, but after splitting up it was just a matter of copying the right config files.
4. It's cleaner. Which IMO is reason enough ;)

@eddiemonge yml? No, just no! ;) The benefit of having the configs as seperate modules is that you can write JS in them: declare task config specific functions, load other files, etc.
Just curious: how is json easier to read/edit/comprehend? The config files nearly are json except for the `strict` and `module.exports` statements

mulderp:
I have seen the practice in the https://github.com/walmartlabs/generator-thorax - and I liked that since, e.g. a connect task is not much related to a uglify or Sass task. What is your experience, @trombom or @eastridge  ?

addyosmani:
@creynders one idea I'd like to suggest is publishing your PR as a separate generator and authoring a blog post about a split-file config version, asking for the community to add their feedback about it to this issue. I'd be happy for us to canvas for further opinions via that route.

We all have different preferences and patterns we like to use when structuring projects and we don't want to seem ungrateful at all for your work here. It's important however to understand that if we went down this path, it would have serious implications for many of our projects as:

We would have to work to port all of our official generators over to using this pattern. webapp, although not the most popular generator, is used as a reference source for project-wide changes and many community generators use it in that way too. This change would mean buying into split-task config for all of our supported generators to minimize upstream cost of updates. It's a non-trivial amount of work.

A safer bet is confirming with the community over a period of time that there's enough interest in a split-config version and proving this to the project. As I mentioned, we aren't dead-set in our ways but require proof that this is something with sufficient interest to warrant the project cost of bringing split-file config into all of our generators.

creynders:
> This change would mean buying into split-task config for all of our supported generators to minimize upstream cost of updates.

Ah, hadn't thought about that.  It does touch an inherent problem with yeoman generators: the downstream syncing. It's hard.  Sorry, don't mean to be dissing yeoman, I really like it, but it still feels rough on the edges.

> we aren't dead-set in our ways but require proof that this is something with sufficient interest to warrant the project cost

Understood!

> I'd like to suggest is publishing your PR as a separate generator and authoring a blog post about a split-file config version

Good idea, I withheld from this since literally all of the ideas and code is coming from other people and it somehow feels ... awkward, but I guess I should just stick a big, loud disclaimer at the top.

addyosmani:
For sure! We all care deeply about smoothing those rough edges out. Do let us know if you would like a review of your blog post when you draft it :)

eastridge:
@mulderp I like having things split up into multiple files

eriktrom:
Note - I helped with implementing this idea in generator-thorax but stole the idea from @stefanpenner in https://github.com/stefanpenner/ember-app-kit - there is a write up about it here http://www.thomasboyt.com/2013/09/01/maintainable-grunt.html by @thomasboyt -- they deserve credit for their hard work

creynders:
For those interested in splitting their grunt task configurations: I turned the configuration-loading part into a node module: [load-grunt-configs](https://github.com/creynders/load-grunt-configs)

It allows you to configure your tasks in several`.json` files and node modules. But maybe the most interesting feature is that it allows you to **split the targets for a single task into separate files**, but also group several task configurations into a single file. 

For example: you can define a `watch` target in a file `lint.js`, but also in `test.js`. I.e. you can group related configuration code into a file, regardless of whether the task is used in other targets as well.

@sindresorhus @addyosmani would this versatility make split configuration files more maintainable in your opinion? Just asking out of interest, not trying to convince you guys.

sindresorhus:
@creynders looks very similar to https://github.com/firstandthird/load-grunt-config, you should probably add something to the readme on how yours differ.

> would this versatility make split configuration files more maintainable in your opinion?

Yes, that helps. As you potentially use it to split out flows instead of just tasks.

creynders:
Typical, the last time I checked (a few months ago) there were no modules with this functionality to be found. Should've guessed they'd pop up in the mean time.

> add something to the readme on how yours differ

Good idea, thanks.

addyosmani:
Just wanted to reiterate I'm looking forward to a blog post about this. Would love us to be able to ask the community what they think within the next few weeks :)

silvenon:
I quickly realized why splitting the tasks into files is a bad idea, because you don't really have a good idea about which tasks you have. I guess the best way of dealing with this issue for now is code folding.

This is a very interesting topic indeed.

creynders:
@silvenon That's why I don't declare my tasks in separate files, but keep them inside the Grunt file. I only move the configuration to separate files. 

And. I haven't tried it out in a real project yet, but I think the "flow"-approach might be very convenient. In my current project at work I've got 32 task configurations, i.e 32 files, but it makes so much more sense to group them logically. E.g. `dev.js`, `test.js`, `build.js`, `deploy.js`, ...

@addyosmani I'm starting on it right away. Hope I'll get it finished today.

silvenon:
Yeah, that's what I meant by "tasks", sorry. Splitting configuration makes it harder to see what you have and how it is configured, especially if you're using load-grunt-tasks.

silvenon:
If you're using Vim, you can use [grunt.vim](https://github.com/mklabs/grunt.vim), which makes the task splitting work great.

We could perhaps provide a flag `--split-tasks`, running a script which would separate tasks into files in the `tasks` folder. That way we can avoid duplication and not be obtrusive about this feature.

What do you say, guys?

creynders:
The blog post on handling large grunt files and call for opinions: http://creynders.wordpress.com/2014/02/10/best-way-to-handle-large-grunt-files/

eddiemonge:
can the split by task files still template inherit from the others or are they completely separated? something, some of the tasks reference other tasks and having duplicate code is usually never good

silvenon:
Tasks usually don't get out of hand concerning line numbers, so I don't see much point in separating that. But configuration quickly grows.

eddiemonge:
I mean in configuration you can do

``` js
copy: {
  files: '<%= yeoman.app %>'
},
htmlmin: {
  files: '<%= copy.files %>'
}
```

Completely arbitrary example but the point is the htmlmin.files reference to the copy files. Can that still be done if the copy and htmlmin config blocks are in different files?

janraasch:
@eddiemonge yes, that's not a problem.

I quiet like this. For a small project like the one scaffolded by `generator-webapp` I feel like having **one** `config/tasks.json` with all the basic configuration already helps slimming down the `gruntfile.js` a lot, and does not add to much overhead keeping everything in sync.

janraasch:
In case you'll be scanning this for :+1:. Here's my consent :)

nessthehero:
I see both sides of this.

I too hate having gigantic grunt files, but I totally see how having a directory of files may get confusing and cumbersome. If I had to pick, I'd go for the multiple files because it does ultimately solve the focus problem, but I'd go for a better configuration syntax to reduce space of each task.

Maybe a combination of the two? Separate files for the default settings but use the main gruntfile for overrides? Or is that just defeating the point?

Not to turn the discussion away from Grunt, but I look at Gulp's syntax and it looks a lot slimmer than Grunts. Maybe migrating to a better syntax (whether it be similar to Gulp's or not) is the ultimate better solution.

silvenon:
Yeah, this is more of a problem with Grunt itself, it does seem pretty verbose. Constructing a Gruntfile manually always takes me a while... When I take a look at the Gulpfile in generator-gulp-webapp, it's like 5x shorter than the Gruntfile here. But this is certainly not a discussion for this project :)

silvenon:
Nevertheless, it is really important that the community picks one, and we should talk about these things. (Somewhere else.)

creynders:
@nessthehero 

> Separate files for the default settings but use the main gruntfile for overrides?

Yeah, that's what I've been mulling over with my module too, but I can't seem to decide what the override behaviour should be, i.e. do you override on a task, target or property level?
In other words, if I have a `jshint` config obj in my grunt file does it override the entire `jshint` obj in the separate file? Or does it only override if a target matches? I guess a merge would make most sense...?

@silvenon 

> Yeah, this is more of a problem with Grunt itself, it does seem pretty verbose. 

I agree. I added the possibility to declare task hierarchy at the topmost key to my module, which does seem to shorten things a bit, e.g.: 

``` javascript
{
    "jshint:tests:src": ["tests/*.js"]
}
```

Maybe it's an idea to suggest that as standard (optional) behaviour in grunt? It has two added benefits: you can easily search for `<task>:<target>` and it's far easier to move a partial configuration object from one file to another.

crrobinson14:
I work with a team of several developers and we work as much on our build process automation as on our code. Splitting out tasks and configs minimizes the chance of merge conflicts. I think for an Open Source project I can see the Yeoman point of view, but in any other environment splitting makes a lot of sense to me.

As for overrides, I'm not sure what they're supposed to accomplish. Grunt tasks have defaults - in the modules that provide the tasks themselves. The only real purpose of the config is to override them, and a lot of Grunt files are bigger than they need to be because people set options to the same values as the defaults. Do you really spend much time overriding these overrides? (Not criticizing - I'm actually curious to know if there are workflows where this is important.)

silvenon:
I guess they're setting it to their default values to make sure the values stay if the default ones change.

Splitting tasks into multiple files works well if you have a good system of navigating through them, like  the Vim plugin I linked. Unfortunately, I didn't find a Sublime plugin which does something similar... Anyway, having tasks split into files can sort of hide them and you can lose control over what you have. Though if you tried that method for a while and it really works for you, I guess it's only inconvenient for some people. But without the Vim plugin I found, I wouldn't split them either.

creynders:
@crrobinson14 I see only one valid use case for the overrides: temporarily spiking a change to a build process.

crrobinson14:
@silvenon I don't understand what this Vim plugin is. I use Komodo and WebStorm and both have really good file navigators (WebStorm also has a fast-open). But even if they didn't, when you break them out each Grunt config is just a file with the same name as its module. I have 16 of them and don't find it difficult to navigate at all. It's certainly much easier than if they were all in a single file. Compared to the rest of my code base, this is the simplest collection there is.

Another thing I love about the split is that Grunt today really does two things: it has modules that provide functions, and tasks that use/combine them in various ways. A typical Gruntfile is full of these two things. Refactoring them out leaves a fairly empty Gruntfile, and all that space makes me wonder - what else can we do here? It would be really interesting to see some more sophistication related to pipelining inputs and outputs between tasks, exposing build process meta-data in structured ways to external processes, etc.

Oh and @creynders good point about the spikes. I'll have to think that one over.

silvenon:
Oh, I guess I never linked it. It's actually similar to what you described. Let's say you have you configuration split in files, `tasks/usemin.js`, `tasks/connect.js` etc. Then you can go to `Gruntfile.js`, place your cursor over `connect:livereload` and press `gf`, this will jump to `tasks/connect`. The plugin helps you bring these split configuration files closer, even though they're split.

Though it seems either Gulp will be taking over the task runner world or Grunt 1.0 will drastically change and become a valid alternative again. I guess we shouldn't waste **too** much time thinking about how to deal with Grunt's configuration syntax.

silvenon:
Check out [this](http://markdalgleish.github.io/presentation-build-wars-gulp-vs-grunt/) presentation.

creynders:
Just an FYI/shameless plug for the split-config file enthusiasts: I wrote a command-line util which will take your gruntfile and use it to generate the separate config files automatically: [grunt-generate-configs](https://github.com/creynders/grunt-generate-configs)

0gust1:
FYI : 

I hacked a kind of Grunt boilerplate for that kind of purpose. 
- Basically, it merges config fragments and does require() on tasks. 
- You can split or organize your conf and tasks files the way you want, it's only the file suffix which tells if it's a config file or a task file.
- So, you can split and organize the config or files the way you want : _by feature_ (that's what I needed), by technical layer, by environnement, etc.

See : https://github.com/0gust1/grunt-organize-boilerplate#why-

I did'nt made a plugin or module, because I don't feel polluting npm for such a little utility.

PR #246, opened 2014-01-21
Pass target through from deprecated server task

callumlocke:
So `grunt server:dist` does not incorrectly proxy to `grunt serve`.


addyosmani:
LGTM. Thanks.

PR #250, opened 2014-01-27
output modernizr to script/vendor folder

### LINKED ISSUE #249: why is the modernizr in dist /bower_components instead of  dist/scripts?
I'm a bit new with yeoman,  Finished a website and been using 'grunt' to make a 'final' version.

 I've been looking at the dist folder : I thought that when using the 'grunt' command alone a distribution version of the site was compiled, and that components were minified and placed in dist/scripts or dist/scripts/vendor and that there would be no trace of the bower_components folder.. but it is there with only modernizr in it...  Is that intended or is it a bug? 

In other words: why is there a bower_components folder in dist for modernizr? shouldn't all scripts be in /scripts and all css in /styles ,,.. ?

eddiemonge:
Fixes #249

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -362,7 +362,7 @@ module.exports = function (grunt) {
         // reference in your app
         modernizr: {
             devFile: '<%%= yeoman.app %>/bower_components/modernizr/modernizr.js',
-            outputFile: '<%%= yeoman.dist %>/bower_components/modernizr/modernizr.js',
+            outputFile: '<%%= yeoman.dist %>/scripts/vendor/modernizr.js',
             files: [
                 '<%%= yeoman.dist %>/scripts/{,*/}*.js',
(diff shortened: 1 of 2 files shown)


robwierzbowski:
What do you think about outputting a full development version to .tmp (so that it's never versioned), and an optimized, crawled version to dist?

eddiemonge:
the full version shouldn't be versioned as its managed by bower. The optimized version has always gone to dist. This only changes the path to where that new file is located.

robwierzbowski:
The Bower package isn't official, and won't necessarily be in sync with the official grunt-modernizr version. I've had issues with this in the past (https://github.com/Modernizr/Modernizr/issues/1165).

This could be a separate issue. I'd like to make outputting a dev version a standard option in grunt-modernizr, but right now I think you have to feed it a complete configuration file.

addyosmani:
LGTM.

PR #253, opened 2014-01-30
Switch to official port of Bootstrap to Sass

silvenon:
@jlong [recommended](https://github.com/jlong/sass-bootstrap/blob/master/README.md#bootstrap--) that we use the official port in the future.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -122,7 +122,7 @@ AppGenerator.prototype.writeIndex = function writeIndex() {
 
   // wire Twitter Bootstrap plugins
   if (this.includeBootstrap) {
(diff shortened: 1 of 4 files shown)


silvenon:
Bootstrap's `bootstrap.js` gets imported by `bower-install`, which is unwanted. I could just ignore it via the `exclude` option, but how does `bower-install` know not to include Modernizr?

stephenplusplus:
> how does bower-install know not to include Modernizr?

If a file is [already referenced in the HTML](https://github.com/yeoman/generator-webapp/blob/master/app/templates/index.html#L21), `grunt-bower-install` won't import it.

silvenon:
Oh, that's very cool. Then `exclude` it is.

addyosmani:
Changes LGTM.

PR #254, opened 2014-02-01
Update time-grunt

silvenon:
`time-grunt` had a bug (sindresorhus/time-grunt#29) caused by early
exit, which broke some useful stuff like `grunt --help`.


sindresorhus:
not needed. ~ means it will get all patch releases.

reading material: https://github.com/isaacs/node-semver

silvenon:
Cool feature :)

PR #260, opened 2014-02-03
Update to Apache Server Configs v2.2.0

alrra:
:bowtie:

### DIFF
--- a/app/templates/htaccess
+++ b/app/templates/htaccess
@@ -1,4 +1,4 @@
-# Apache Server Configs v2.1.0 | MIT License
+# Apache Server Configs v2.2.0 | MIT License
 # https://github.com/h5bp/server-configs-apache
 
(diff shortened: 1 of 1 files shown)


app/templates/htaccess
```diff
@@ -258,14 +265,14 @@ AddDefaultCharset utf-8
 # | Clickjacking                                                               |
 # ------------------------------------------------------------------------------
 
-# Protect web site against clickjacking.
+# Protect website against clickjacking.
```
sindresorhus, line 268: important stuff

alrra, line 268: @sindresorhus believe it or not, someone complained about that :)

sindresorhus, line 268: When a project gets popular enough everything is under the microscope ;)

hemanth, line 268: :+1:

Issue #259, opened 2014-02-03
BUG: copy task config output unprocessed


creynders:
I'm pretty certain this is a bug with the current version, after scaffolding the `copy` task `src` looks like this:

``` javascript
src    : [
    '*.{ico,png,txt}', '.htaccess', 'images/{,*/}*.webp', '{,*/}*.html', 'styles/fonts/{,*/}*.*',
    'bower_components/' + (this.includeCompass
    ? 'sass-'
    : '') + 'bootstrap/' + (this.includeCompass
    ? 'fonts/'
    : 'dist/fonts/') + '*.*'
]
```

@silvenon could it be this is the result of your PR's?

silvenon:
Yes, it could be, but I'm pretty sure that was a bug before my PR too. If you check the diff of my PR, you'll see that I just used the same coding style. Or perhaps I did make a mistake.

Anyway, I'll fix this in a few hours.

eddiemonge:
should be fixed now

PR #261, opened 2014-02-03
Fix Bootstrap font paths

silvenon:
This fixes the wrong output in the Gruntfile concerning Bootstrap fonts (#259). One question, though, what is this variable for?

``` scss
$icon-font-path: "/bower_components/bootstrap-sass/vendor/assets/fonts/";
```

It seems completely useless because there's already a Bootstrap partial which imports the Glyphicons. May I delete it?

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -345,7 +345,9 @@ module.exports = function (grunt) {
                         'images/{,*/}*.webp',
                         '{,*/}*.html',
                         'styles/fonts/{,*/}*.*'<% if (includeBootstrap) { %>,
-                        'bower_components/bootstrap' + (this.includeCompass ? '-sass/' : '/') + (this.includeCompass ? 'vendor/assets/fonts/bootstrap/' : 'dist/fonts/') +'*.*'<% } %>
+                        <% if (includeCompass) {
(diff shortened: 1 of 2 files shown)


app/templates/Gruntfile.js
```diff
@@ -345,7 +345,9 @@ module.exports = function (grunt) {
                         'images/{,*/}*.webp',
                         '{,*/}*.html',
                         'styles/fonts/{,*/}*.*'<% if (includeBootstrap) { %>,
-                        'bower_components/bootstrap' + (this.includeCompass ? '-sass/' : '/') + (this.includeCompass ? 'vendor/assets/fonts/bootstrap/' : 'dist/fonts/') +'*.*'<% } %>
+                        <% if (includeCompass) {
+                        %>'bower_components/bootstrap-sass/vendor/assets/fonts/bootstrap/*.*'<% } else {
```
eddiemonge, line 349: can you follow the existing code convention of having the closing template tag on the same line as the opening please?

silvenon, line 349: What do you propose? Writing all of that on the same line? I thought maybe this would be more readable.

I couldn't make a line break inside the tag because that line break would be outputted.

PR #262, opened 2014-02-04
Add Gemfile when Sass & Compass is selected

### LINKED ISSUE #257: Gemfile
Perhaps a `Gemfile` could be created and bundled for the users who choose Sass & Compass.

silvenon:
The generator creates a Gemfile and runs `bundle install` to ensure Sass and Compass are installed (if Sass & Compass option is selected). This closes #257.

I also updated the tests to check whether the Gemfile is created, but test seem kinda weird. I never tested before, but it seems they don't test all combinations and the file is named `test.js` instead of `test.spec.js`, which would be more suitable for jasmine-node. Also, when the tests fail, the testing just stops, it does tell you how many tests you ran and how many of them failed. Plus, the output says that there are 0 assertions, which is not true, and that's because the tests aren't using Jasmine's assertions. Am I running the tests wrong, or should I open a new issue?

There's also a `bootstrap-sass` gem, but I didn't really know what to do with it concerning Glyphicons, so I just left it out of the Gemfile and decided to stick with Bower.


eddiemonge:
This seems kind of mean to kick them out of the generator if they choose compass/sass and they dont have it installed. It should just warn them at the end. 

I'm also not sure about running `bundle install` either. What if they want to install a different version or from a different location?

silvenon:
I'm not kicking them out of the generator, on the contrary, I'm **ensuring** that they'll have it once the generator is done. If they don't have Sass & Compass, those dependencies will be installed for them.

I'm kicking them out if they don't have Ruby and Bundler installed, though. Should I let it pass? Also, I should provide info on how to install those.

Do you guys agree I should point them to [rbenv](https://github.com/sstephenson/rbenv) for installing Ruby? It's the best one, and [Sam Stephenson](https://github.com/sstephenson?tab=repositories) made it, which speaks by itself.

What if they want a different version of grunt-autoprefixer? They'll just change it, right? :) What do you mean by "different location"? Like not RubyGems?

eddiemonge:
doesnt https://github.com/yeoman/generator-webapp/pull/262/files#diff-2018087f584c4398b5c3a23fc0e5f9dbR92 exit the generator?

As for the different versions, I don't mean the version number. I mean the normal installed one when running `gem install` vs a locally made one with `gem install --local`.

Also how does this work with things like rvm, rbenv, or brew installed ruby, etc. What decides which ruby/gem to use if the .ruby-version hasn't been created yet. I think something like this is outside of the scope of Yeoman and merely checking for the existence of the gems is more then enough, and might even be too much. If it was in the JS world, like say with the libsass port to nodeJS it would be easier and more understandable for us to include here.

silvenon:
Ruby version, that's a good point. We can use their global Ruby version, I don't think there's anything wrong with that. (I edited my comment above, I didn't notice you already commented.) Do you think I should add a `.ruby-version` corresponding to their global Ruby version and use that version in the `Gemfile` too?

Actually, I don't know what that error part does, but it doesn't stop the generator (I tried now). I will remove that.

eddiemonge:
I don't think using the global Ruby version is good because it could be grossly out of date. This is what I mean by I think the Ruby side of things is out of scope. There are just way too many variables. Maybe it might be better just to link them to some resources on getting started with Ruby/Sass. Or better yet, remove Compass completely and switch to node-sass so the whole Ruby issue is sidestepped. 

@addyosmani @sindresorhus @passy thoughts on Ruby/Sass/Compass?

silvenon:
If they use Sass regularly, which I'm sure they do, that's why they chose the option, they already have Ruby set up. It would be very weird if they're setting up a local Ruby version for each project, while having a horribly outdated global version.

But yeah, I'm all for node-sass, but I think libsass is a bit [covered with issues](https://github.com/hcatlin/libsass/issues?state=open). Those are mostly related to new features, but still...

passy:
I wish I had a good advice here, but unfortunately whether it's a single system-wide Ruby+Gems installation, RVM or bundler, it's always been a very unpleasant experience for me, both on OS X and Linux. If we could get rid of Ruby as dependency altogether, I'd definitely support that 100%.

My recent experience with libsass [was really positive](https://twitter.com/passy/status/428298994306252800), but the maintainers themselves [don't recommend using it in production at this point](https://twitter.com/sindresorhus/status/430740165875605504).

silvenon:
RVM is a pretty rough experience, yes, I wanted to uninstall it while I was installing it, I felt like it was hacking my computer. We can make a simple recommendation, we won't (and can't) act upon which Ruby management system they use, that's user's choice.

What is the problem? RVM vs rbenv?

janko:
As I understand, this generator depends on Ruby. Maybe the process of generating doesn't, but everything else does, because Sass & Compass are Ruby libraries.

@silvenon proposes adding a `Gemfile` to the project, to add the ability for users to specify versions of Sass and Compass. This ability is arguably a must-have. About different choosing locations of gems, it is very rare, and in that case they can easily achieve that by specifying the version of the installed gem. If you would rather go without Ruby, that's probably a discussion for a different topic ;)

A check if Ruby/Bundler is installed should be added, with the appropriate installation links. And if they have Ruby installed already, whether they used RVM, Rbenv or Chruby is irrelevant, these projects abstract themselves away.

silvenon:
I decided to only warn them if they don't have Compass installed. Is this okay with everyone?

silvenon:
Actually, not even that. I completely changed my mind about these changes. I'm betting this project will eventually use grunt-sass (libsass) anyway.

PR #263, opened 2014-02-04
install bower_components in root directory

### LINKED ISSUE #171: Don't install bower components in /app
This is just a proposal.

I think bower components should be installed in the root directory or in some kind of /vendor or /lib folder (more on why in a second). The reason I think it should not be installed in /app is that folder should be reserved for the user's app code, meaning that anything in there is maintainable by the user or created by the user. Bower should be managing the b_c folder, not the user.

It would be nice if npm supported installing to a named directory so b_c and node_modules could both be under a single folder.

B_c files could be supported with middleware in connect proxy similar to how .tmp is done so it wouldn't affect dev. This means that it would need to be in a subfolder then so it could be served as /bower_components still as if you point the middleware to <% bower_components %> it will try to serve from a b_c folder there instead of the actual folders (unless there is an option I am missing).

eddiemonge:
/app should only have user created/editable files. Bower should install
its files alongside npm's.

Fixes #171

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -15,15 +15,17 @@ module.exports = function (grunt) {
     // Time how long tasks take. Can help when optimizing build times
     require('time-grunt')(grunt);
 
+    // Configurable paths
+    var config = {
+        app: 'app',
+        dist: 'dist'
+    };
+
     // Define the configuration for all the tasks
     grunt.initConfig({
 
         // Project settings
-        yeoman: {
-            // Configurable paths
-            app: 'app',
-            dist: 'dist'
-        },
+        yeoman: config,
 
         // Watches files for changes and runs tasks based on the changed files
         watch: {<% if (coffee) { %>
@@ -74,32 +76,38 @@ module.exports = function (grunt) {
         connect: {
             options: {
                 port: 9000,
+                open: true,
                 livereload: 35729,
                 // Change this to '0.0.0.0' to access the server from outside
                 hostname: 'localhost'
             },
             livereload: {
                 options: {
-                    open: true,
-                    base: [
-                        '.tmp',
-                        '<%%= yeoman.app %>'
-                    ]
+                    middleware: function(connect) {
+                        return [
+                            connect.static('.tmp'),
+                            connect().use('/bower_components', connect.static('./bower_components')),
+                            connect.static(config.app)
+                        ]
+                    }
                 }
             },
             test: {
                 options: {
+                    open: false,
                     port: 9001,
-                    base: [
-                        '.tmp',
-                        'test',
-                        '<%%= yeoman.app %>'
-                    ]
+                    middleware: function(connect) {
+                        return [
+                            connect.static('.tmp'),
+                            connect.static('test'),
+                            connect().use('/bower_components', connect.static('./bower_components')),
+                            connect.static(config.app)
+                        ]
+                    }
                 }
             },
             dist: {
                 options: {
-                    open: true,
                     base: '<%%= yeoman.dist %>',
                     livereload: false
                 }
@@ -185,7 +193,7 @@ module.exports = function (grunt) {
                 imagesDir: '<%%= yeoman.app %>/images',
                 javascriptsDir: '<%%= yeoman.app %>/scripts',
                 fontsDir: '<%%= yeoman.app %>/styles/fonts',
-                importPath: '<%%= yeoman.app %>/bower_components',
+                importPath: './bower_components',
                 httpImagesPath: '/images',
                 httpGeneratedImagesPath: '/images/generated',
                 httpFontsPath: '/styles/fonts',
@@ -224,13 +232,12 @@ module.exports = function (grunt) {
             app: {
                 src: ['<%%= yeoman.app %>/index.html'],
                 ignorePath: '<%%= yeoman.app %>/'<% if (includeCompass) { %>,
-                exclude: ['<%%= yeoman.app %>/bower_components/bootstrap-sass/vendor/assets/javascripts/bootstrap.js']
+                exclude: ['./bower_components/bootstrap-sass/vendor/assets/javascripts/bootstrap.js']
             },
             sass: {
                 src: ['<%%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
-                ignorePath: '<%%= yeoman.app %>/bower_components/'
-            }<% } else { %>
-            }<% } %>
+                ignorePath: '<%%= yeoman.app %>/bower_components/'<% } %>
+            }
         },
 
         // Renames files for browser caching purposes
@@ -349,11 +356,18 @@ module.exports = function (grunt) {
                         '.htaccess',
                         'images/{,*/}*.webp',
                         '{,*/}*.html',
-                        'styles/fonts/{,*/}*.*'<% if (includeBootstrap) { %>,<% if (includeCompass) { %>
-                        'bower_components/bootstrap-sass/vendor/assets/fonts/bootstrap/*.*'<% } else { %>
-                        'bower_components/bootstrap/dist/fonts/*.*'<% } %><% } %>
+                        'styles/fonts/{,*/}*.*'
                     ]
-                }]
+                }<% if (includeBootstrap) { %>, {
+                    expand: true,
+                    dot: true,
+                    cwd: '.',
+                    dest: '<%%= yeoman.dist %>',
+                    src: [<% if (includeCompass) { %>
+                        'bower_components/bootstrap-sass/vendor/assets/fonts/bootstrap/*.*'<% } else { %>
+                        'bower_components/bootstrap/dist/fonts/*.*'<% } %>
+                    ]
+                }<% } %>]
             },
             styles: {
                 expand: true,
@@ -367,7 +381,7 @@ module.exports = function (grunt) {
         // Generates a custom Modernizr build that includes only the tests you
         // reference in your app
         modernizr: {
-            devFile: '<%%= yeoman.app %>/bower_components/modernizr/modernizr.js',
+            devFile: './bower_components/modernizr/modernizr.js',
             outputFile: '<%%= yeoman.dist %>/scripts/vendor/modernizr.js',
             files: [
                 '<%%= yeoman.dist %>/scripts/{,*/}*.js',
--- a/app/templates/bowerrc
+++ b/app/templates/bowerrc
@@ -1,3 +1,3 @@
 {
-    "directory": "app/bower_components"
+    "directory": "bower_components"
 }
--- a/app/templates/gitignore
+++ b/app/templates/gitignore
@@ -2,5 +2,5 @@ node_modules
 dist
 .tmp
 .sass-cache
-app/bower_components
+bower_components
 test/bower_components


app/templates/Gruntfile.js
```diff
@@ -74,32 +76,38 @@ module.exports = function (grunt) {
...
-                    open: true,
-                    base: [
-                        '.tmp',
-                        '<%%= yeoman.app %>'
-                    ]
+                    middleware: function(connect) {
+                        return [
+                            connect.static('.tmp'),
+                            connect().use('/bower_components', connect.static('./bower_components')),
+                            connect.static('app')
```
silvenon: This should also be `config.app`.

silvenon:
I think all `bower_components` links need to be preceded with `/` (`/bower_components`) to work in HTML files which are not in the root.

eddiemonge:
thats a different issue and not related to this pr

silvenon:
Whoops, you're right, I don't know why I said that here.

eddiemonge:
@sindresorhus @addyosmani @passy look good?

app/templates/Gruntfile.js
```diff
@@ -224,13 +232,12 @@ module.exports = function (grunt) {
             app: {
                 src: ['<%%= yeoman.app %>/index.html'],
                 ignorePath: '<%%= yeoman.app %>/'<% if (includeCompass) { %>,
```
silvenon, line 234: Is this necessary now that that there will be no `app/` in the path? (Same goes for the `yeoman.app` bit in the second `ignorePath`.)

kevva, line 234: Shouldn't be necessary, no.

silvenon:
I've been using this branch for a while and it works fine. It would be nice to see this merged soon.

silvenon:
Actually it doesn't. Bower components don't get linked properly in the build process.

``` sh
$ mkdir webapp-test
$ cd webapp-test
$ yo webapp # leave the defaults
$ bower install --save normalize-css # just to throw something in there
$ grunt bowerInstall:app
$ grunt build
```
- `vendor.css` doesn't even get revved
- `vendor.js` and `plugins.js` get revved, but are empty

silvenon:
This happens because the assets are searched for relative to `index.html`, while they should be searched for from the root. One way is to set that explicitly:

``` html
<!-- build:css(./) styles/vendor.css -->
<!-- build:js(./) scripts/vendor.js -->
<!-- build:js(./) scripts/plugins.js -->
```

Maybe there's a better way, but this works.

app/templates/Gruntfile.js
eddiemonge, line 234: tbh, im not sure what the correct thing for this section is. i should go read the docs on it again but if someone wants to suggest an alternative config, im open to change it

addyosmani:
@eddiemonge do you think you'll be able to address the issues @silvenon mentioned above? I've ran into them too and think that we'll need to ensure builds correctly work in all common use-cases before we can land this.

app/templates/bowerrc
```diff
@@ -1,3 +1,3 @@
 {
-    "directory": "app/bower_components"
+    "directory": "bower_components"
```
haydenbleasel, line 2: You could just delete the `bowerrc` file altogether.

silvenon, line 2: @haydenbleasel I think it's better not to, the configuration relies on it. If someone specified a different folder in their global `.bowerrc` (if such a thing doesn't exist yet, it might eventually), there would be trouble :)

eddiemonge:
oh usemin. causing more problems. Ill take a look

app/templates/Gruntfile.js
silvenon, line 234: This is the proposed config:

``` js
// Automatically inject Bower components into the HTML file
bowerInstall: {
  app: {
    src: ['<%= yeoman.app %>/index.html'],
    exclude: ['bower_components/bootstrap-sass/vendor/assets/javascripts/bootstrap.js']
  },

  sass: {
    src: ['<%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
    ignorePath: 'bower_components/'
  }
}
```

`ignorePath` option removes the specified string from the include path. Previously `bower_components` was in the `app` folder, which would by itself make grunt-bower-install output paths like these:

``` html
<!-- app/index.html -->
<script src="app/bower_components/jquery/jquery.js"></script>
```

But in this case `app/index.html` would be searching for that file in `app/app/bower_components`. To get rid of the `app/` part, you specify it like this: `ignorePath: 'app/'`

addyosmani:
This will need to be rebased whenever we get a chance.

addyosmani:
@eddiemonge just re-pinging in case you're still interested in pursuing this one.

eddiemonge:
I am, I just need to find the time to do it. Maybe later today, unless @silvenon wouldn't mind helping out

silvenon:
Sure, how? Do I make a PR to your fork?

eddiemonge:
or you could merge the code then submit it as a new PR here.

silvenon:
Ok, I'll do the latter.

robwierzbowski:
Even though this is in now, just want to register some concerns on whether or not it makes Yeoman structure more understandable.

There's nothing special about project root that says "don't edit these files". The root is for project workflow and settings code: rc files, things that run our Gruntfile, things that don't end up in the final app. npm_packages belongs here because they run at this level. We don't edit them because we know they're packages, but of course we should edit the settings and other project workflow level files. 

The app directory is for code that ends up in the final app (or code transformed into something that ends up in the final app). Bower packages end up in our final app. I don't see any pattern that moving them to root would standardize. We're not serving anything else from root — IMO breaking / obfuscating our serve and build paths is a usability regression. Also adds some complication to our complex Gruntfile, but that would be fine if it removed complexity elsewhere.

silvenon:
Yeah, this issue is a bit tough. I agree it seems to introduce complexity to Gruntfile without a visible gain, but I like having Bower components at the root because:
- they are automatically generated, `app` doesn't feel like a good place to put generated files
- that's where they are downloaded by default :P
- most of the files from there don't end up in the build
- when `tree`ing the `app` folder, we don't have to ignore `bower_components` anymore :smiley:

robwierzbowski:
I think the naturalness of components for app, served from app, built into dist is more important for user understanding. I'm most concerned with a user saying "I need to reference bower `script-a.js` in head. Where is it? Ah, right, it's in root. Why didn't `../script-a.js`. work? Oh, if I look in the Gruntfile (and understand config's api), or if I read docs, I see that those act like they're symlinked in there." 

It adds complexity and something to remember to a very common task, and one that could be done by any member of a team. On my team I have 15 people that can touch scripts and HTML, but only four or so that have any node or grunt experience. This change creates a barrier to understanding for the majority.

silvenon:
Interesting points... It can bring some confusion, but I think `bower_components` should really be abstracted away, so it doesn't matter where they are, the `bowerInstall` task should be the one doing most of the work.

Also, if you check the updated version of this PR (#310), you'll notice that the links are indeed `../bower_components/script-a.js`, so it's not really that confusing. Components included via `bowerInstall` should hint how to manually include the rest of the components which are not setup properly and can't be included automatically with grunt-bower-install.

robwierzbowski:
Stepping out of the app folder for assets used in app is even worse. The
bower install task is cool, but last time I checked is only good for
automatically adding a 'main' bower file. If you need to add a CSS file
along with the main JS (which we often do), or when you're managing fonts
and images with bower (which we often do), insertion can't be automatic.

On Friday, March 21, 2014, Matija Marohnić notifications@github.com wrote:

> Interesting points... It can bring some confusion, but I think
> bower_components should really be abstracted away, so it doesn't matter
> where they are, the bowerInstall task should be the one doing most of the
> work.
> 
> Also, if you check the updated version of this PR (#310https://github.com/yeoman/generator-webapp/pull/310),
> you'll notice that the links are indeed ../bower_components/script-a.js,
> so it's not really that confusing. Components included via bowerInstallshould hint how to manually include the rest of the components which are
> not setup properly and can't be included automatically with
> grunt-bower-install.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/yeoman/generator-webapp/pull/263#issuecomment-38271255
> .

## 

Rob Wierzbowski
@robwierzbowski http://twitter.com/#!/robwierzbowski
http://github.com/robwierzbowski
http://robwierzbowski.com

robwierzbowski:
Sorry, didn't see at first you said 'hint', not 'replace' :).

My point is we're adding much complexity and new, yeoman only patterns for
minimal gain. The only argument I've seen with merit is "things in app are
editable, and bower_components aren't", but things in root are also
editable. I watch my co-workers use yeoman, and watch them learn and
question the conventions. They've never had trouble with not editing bower
components, just like they never edit npm packages. I worry that this takes
us further from a natural project structure and adds our own magic that
each new yeoman user will have to learn. As a generator author I know that
I'll have to adopt these patterns for composability with other generators
and I'm reluctant because I think they're a usability regression, and a
large part of what I do is teach people how to use my generator and yeoman
in general.

On Friday, March 21, 2014, Rob Wierzbowski rob.wierzbowski@gmail.com
wrote:

> Stepping out of the app folder for assets used in app is even worse. The
> bower install task is cool, but last time I checked is only good for
> automatically adding a 'main' bower file. If you need to add a CSS file
> along with the main JS (which we often do), or when you're managing fonts
> and images with bower (which we often do), insertion can't be automatic.
> 
> On Friday, March 21, 2014, Matija Marohnić <notifications@github.com<javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
> 
> > Interesting points... It can bring some confusion, but I think
> > bower_components should really be abstracted away, so it doesn't matter
> > where they are, the bowerInstall task should be the one doing most of
> > the work.
> > 
> > Also, if you check the updated version of this PR (#310https://github.com/yeoman/generator-webapp/pull/310),
> > you'll notice that the links are indeed ../bower_components/script-a.js,
> > so it's not really that confusing. Components included via bowerInstallshould hint how to manually include the rest of the components which are
> > not setup properly and can't be included automatically with
> > grunt-bower-install.
> > 
> > —
> > Reply to this email directly or view it on GitHubhttps://github.com/yeoman/generator-webapp/pull/263#issuecomment-38271255
> > .
> 
> ## 
> 
> Rob Wierzbowski
> @robwierzbowski http://twitter.com/#!/robwierzbowski
> http://github.com/robwierzbowski
> http://robwierzbowski.com

## 

Rob Wierzbowski
@robwierzbowski http://twitter.com/#!/robwierzbowski
http://github.com/robwierzbowski
http://robwierzbowski.com

silvenon:
I agree that the Gruntfile configuration is a bit confusing with all the middleware options, but the end result is actually pretty expected. Stepping out of the app folder is intuitive, because `bower_components` is located at `../`. I don't think it's that hard to figure out the paths for those additional CSS files, images and fonts, they only have one extra `../`, which totally makes sense. Fonts are probably the trickiest to include, but I don't really believe we often include them via Bower, we usually use Font Squirrel, but most often we use CDNs like Google Fonts, Typekit etc. Also, I'm sure Bower will get better and grunt-bower-install will eventually be able to automatically include those files as well. Also, I think of `app` and `bower_components` as `app` and `vendor` folders in Rails, so I'm kinda biased.

I don't really see how this adds complexity, other than the configuration of the `connect` task. One thing I noticed is that now we have `bower_components` in the root and in the `test` directory. It was more obvious when `bower_components` was in the `app` directory, I didn't notice that before.

I understand that generator-webapp is a collection of best practices and I agree that we should put more thought in changing the project's structure and that this solution isn't perfect. @eddiemonge, @addyosmani and @sindresorhus, what do you guys think?

robwierzbowski:
I'm not a rails guy so that pattern isn't super familiar to me. Totally respect your opinions, these decisions are hard. Also sorry I was busy before the PR went in and wasn't able to comment earlier.

brohug 

I can see a benefit if we're installing test and app libs into a single root dir. That's a convincing use case.

silvenon:
:metal:

eddiemonge:
well considering it was my proposal I'm totally for it :accept: :100: 

If we can move the `/test` bower components to the root as well that would be awesome as I troubleshooted a bug yesterday where the person didn't know they were supposed to do `bower install` from the test directory as well.

At the end of the day, Yeoman is an opinionated scaffolding tool that is meant to help people get their work done faster. It isn't the end all declarative statement on how user's have to work but merely suggestions that they can take or leave as needed.

robwierzbowski:
> I can see a benefit if we're installing test and app bower libs into a single root dir. That's a convincing use case.

Just realized that this was handled before by having the test connect server cascading from `test` to `.tmp` to `app`.

silvenon:
Perhaps including test Bower components under `devDependencies`? I'll see how that works...

PR #278, opened 2014-02-18
Replace load-grunt-tasks with jit-grunt

creynders:
Jit-grunt is fantastic:

> A JIT(Just In Time) plugin loader for Grunt.

I.e. it only loads the plugins you need for the running tasks. It saves **a lot** of build time. My spikes show me 5x speed improvements.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -10,7 +10,7 @@
 module.exports = function (grunt) {
 
     // Load grunt tasks automatically
-    require('load-grunt-tasks')(grunt);
+    require('jit-grunt')(grunt);
 
     // Time how long tasks take. Can help when optimizing build times
     require('time-grunt')(grunt);
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -10,7 +10,9 @@
 module.exports = function (grunt) {
 
     // Load grunt tasks automatically
-    require('jit-grunt')(grunt);
+    require('jit-grunt')(grunt,{
+        useminPrepare: 'grunt-usemin'
+    });
 
     // Time how long tasks take. Can help when optimizing build times
     require('time-grunt')(grunt);
--- a/app/templates/_package.json
+++ b/app/templates/_package.json
@@ -26,9 +26,9 @@
     "grunt-newer": "~0.6.0",
     "grunt-svgmin": "~0.2.0",
     "grunt-concurrent": "~0.4.0",
-    "load-grunt-tasks": "~0.2.0",
     "time-grunt": "~0.2.0",
-    "jshint-stylish": "~0.1.3"
+    "jshint-stylish": "~0.1.3",
+    "jit-grunt": "~0.2.1"
   },
   "engines": {
     "node": ">=0.8.0"


sindresorhus:
Can you show a number comparison (use time-grunt) before and after when used here?

creynders:
Here's the results: https://gist.github.com/creynders/36830565d392f73d6ec6
With a vanilla install it's about 2 times as fast. But I think the difference becomes bigger, the more plugins you use.
However, good you asked for details, since I noticed it doesn't find the `useminPrepare` task (which makes sense). Pushing the correct config in a sec.

sindresorhus:
I tried it out with a fresh `yo webapp` and couldn't see much difference.

load-grunt-tasks

```
loading tasks                                                             2ms  0%
default                                                                   0ms  0%
newer:jshint                                                              2ms  0%
newer:jshint:all                                                          7ms  0%
jshint:all                                                              126ms  ▇▇ 2%
newer-postrun:jshint:all:1:/Users/...t/node_modules/grunt-newer/.cache    1ms  0%
test                                                                      1ms  0%
clean:server                                                              3ms  0%
concurrent:test                                                         718ms  ▇▇▇▇▇▇▇▇▇ 12%
autoprefixer                                                              1ms  0%
autoprefixer:dist                                                        11ms  0%
connect:test                                                              7ms  0%
mocha                                                                     1ms  0%
mocha:all                                                                1.1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 19%
build                                                                     1ms  0%
clean:dist                                                                7ms  0%
useminPrepare                                                             0ms  0%
useminPrepare:html                                                       27ms  0%
concurrent:dist                                                          2.8s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 49%
autoprefixer                                                              0ms  0%
autoprefixer:dist                                                       113ms  ▇▇ 2%
concat                                                                    0ms  0%
concat:generated                                                         14ms  0%
cssmin                                                                    0ms  0%
cssmin:generated                                                         57ms  ▇ 1%
uglify                                                                    0ms  0%
uglify:generated                                                        412ms  ▇▇▇▇▇ 7%
copy:dist                                                                 9ms  0%
modernizr                                                               320ms  ▇▇▇▇ 5%
rev                                                                       1ms  0%
rev:dist                                                                  5ms  0%
usemin                                                                    0ms  0%
usemin:html                                                              13ms  0%
usemin:css                                                                9ms  0%
htmlmin                                                                   0ms  0%
htmlmin:dist                                                             11ms  0%
Total 5.8s
```

jit-grunt

```
loading tasks                                                             4ms  0%
default                                                                   0ms  0%
loading grunt-newer                                                      11ms  0%
newer:jshint                                                              1ms  0%
newer:jshint:all                                                          6ms  0%
loading grunt-contrib-jshint                                             50ms  ▇ 1%
jshint:all                                                              148ms  ▇▇ 3%
newer-postrun:jshint:all:1:/Users/...t/node_modules/grunt-newer/.cache   71ms  ▇ 1%
test                                                                      0ms  0%
loading grunt-contrib-clean                                               5ms  0%
clean:server                                                              4ms  0%
loading grunt-concurrent                                                  6ms  0%
concurrent:test                                                         282ms  ▇▇▇▇ 5%
loading grunt-autoprefixer                                               37ms  ▇ 1%
autoprefixer                                                              0ms  0%
autoprefixer:dist                                                        15ms  0%
loading grunt-contrib-connect                                            47ms  ▇ 1%
connect:test                                                              7ms  0%
loading grunt-mocha                                                      49ms  ▇ 1%
mocha                                                                     0ms  0%
mocha:all                                                                1.1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 20%
build                                                                     1ms  0%
clean:dist                                                                7ms  0%
loading grunt-usemin                                                     11ms  0%
useminPrepare                                                             0ms  0%
useminPrepare:html                                                       14ms  0%
concurrent:dist                                                          2.6s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 47%
autoprefixer                                                              0ms  0%
autoprefixer:dist                                                       123ms  ▇▇ 2%
loading grunt-contrib-concat                                              2ms  0%
concat                                                                    1ms  0%
concat:generated                                                         21ms  0%
loading grunt-contrib-cssmin                                             19ms  0%
cssmin                                                                    0ms  0%
cssmin:generated                                                         77ms  ▇ 1%
loading grunt-contrib-uglify                                             54ms  ▇ 1%
uglify                                                                    0ms  0%
uglify:generated                                                        393ms  ▇▇▇▇▇ 7%
loading grunt-contrib-copy                                                1ms  0%
copy:dist                                                                 9ms  0%
loading grunt-modernizr                                                   2ms  0%
modernizr                                                               333ms  ▇▇▇▇▇ 6%
loading grunt-rev                                                         1ms  0%
rev                                                                       0ms  0%
rev:dist                                                                  5ms  0%
loading grunt-usemin                                                      1ms  0%
usemin                                                                    0ms  0%
usemin:html                                                              12ms  0%
usemin:css                                                                4ms  0%
loading grunt-contrib-htmlmin                                             4ms  0%
htmlmin                                                                   0ms  0%
htmlmin:dist                                                             11ms  0%
Total 5.6s
```

creynders:
Hmm, not sure why you don't see a difference. Granted the second run seems to be faster, so I did some further tests, for both load-grunt-tasks and jit-grunt I did a clean install, then ran `grunt` one time and subsequently logged 3 more iterations, both for `grunt --verbose` and `grunt test --verbose`. There's a clear difference on my system:

| task | load-grunt-tasks | jit-grunt |
| --- | --: | --: |
| grunt [2] | 11.6s | 7.4s |
| grunt [3] | 11.1s | 7.5s |
| grunt [4] | 10.1s | 7.6s |
| grunt test [2] | 2.9s | 2.1s |
| grunt test [3] | 3s | 2.2s |
| grunt test [4] | 3s | 2.2s |

See [updated gist](https://gist.github.com/creynders/36830565d392f73d6ec6) for outputs.

But most importantly, as I wrote, the differences become greater the bigger the project.

eddiemonge:
It does seem to be consistently fater

shootaroo:
Hi, It is not a correct comparison, because there is a load-grunt-tasks before the time-grunt.
https://github.com/yeoman/generator-webapp/blob/master/app/templates/Gruntfile.js#L12-L16

``` js
    // Load grunt tasks automatically
    require('load-grunt-tasks')(grunt);

    // Time how long tasks take. Can help when optimizing build times
    require('time-grunt')(grunt);
```

Please measure out in reverse order.

jit-grunt is more effective with fewer task executions, e.g. `grunt clean`, `grunt serve`.

creynders:
@shootaroo If I remember correctly I _did_ require time-grunt before load-grunt-tasks or jit-grunt. Maybe that would explain the difference in results with @sindresorhus ' tests?

addyosmani:
@eddiemonge could you share some of the numbers you were getting during comparison testing?

shootaroo:
@creynders @sindresorhus loading time of the plugins by load-grunt-tasks is not included.

``` js
    require('load-grunt-tasks')(grunt);
    require('time-grunt')(grunt);
```

```
loading tasks                                                     3ms
```

time-grunt before load-grunt-tasks on my MacBook Air,

``` js
    require('time-grunt')(grunt);
    require('load-grunt-tasks')(grunt);
```

```
loading tasks                                                     728ms
```

It takes more, depending on the number of plugin and environment.

creynders:
What's the status on this? If anything else needs to be done to convince you guys, let me know.

eddiemonge:
It shaves a small amount of time off my production-ish app: https://gist.github.com/eddiemonge/10015110

eddiemonge:
The only thing I didn't like with jit-grunt was the manual aliasing. It isn't super clear how/when its done when you add new things to a grunt process.

eddiemonge:
@creynders can you rebase and update to the latest version of jit-grunt to see if the manual aliasing is really resolved? It might also be helpful if you had a repo we could test with to see the performance differences

ain:
:+1: for JIT Grunt. (Huge) speed difference comes with LiveReload for instance, in cases where you don't need the whole stack loaded.

eddiemonge:
@sindresorhus jit-grunt is actually far faster. I've been testing it and seeing 10x improvement in some cases. Downside is that for things like usemin or grunt-angular-templates, you have to map them

eddiemonge:
While I do want to land this, I have to instead close it for the reasons I give in: https://github.com/yeoman/yeoman/issues/1382

Thank you for taking the time on this PR and comments. I do like the plugin and will continue to use it in my projects.

nickautomatic:
I wonder if this is worth taking another look at. jit-grunt now seems to be pretty good at finding tasks: in fact, I swapped it out in an existing webapp-based project, and found it actually involved _less_ configuration: I'd had to add a custom pattern to load-grunt-tasks to get it to find "assemble"; jit-grunt found assemble without any custom configuration at all.

eg. I replaced:

```
require('load-grunt-tasks')(grunt, { pattern: ['grunt-*', 'assemble'] });
```

...with:

```
require('jit-grunt')(grunt); 
```

So in this case, having jit-grunt would actually have made it _easier_ to add a custom task (certainly for me - a relative beginner with Grunt - it took quite a lot of head-scratching to get assemble to work with yo webapp; with jit-grunt it would have just worked). jit-grunt now searches for tasks in three different places ("grunt-task", "grunt-contrib-task", and "task"), whereas load-grunt-tasks, as I understand it, only looks for the first of these with its default config.

Looking at jit-grunt's history, I think its task-finding functionality was overhauled in Oct 2014 (after this issue was closed). As far as I can see, it's now not only much faster than load-grunt-tasks but also requires no more - and possibly less - configuration.

eddiemonge:
Still doesn't work for a lot of things I found.

creynders:
The speed gain outweighs the few times you have to add some configuration to jit-grunt by faaaaaaar.

Seriously, it's _very_ ironic IMO that jit-grunt usage is deemed too complex, while the configuration that goes into initConfig is way more complex, opaque and hard to follow.

ain:
We've built things of great magnitude and esp. if you rely on livereload, `jit-grunt` is a must. Configuration is the least thing you should be worried about in that case.

nickautomatic:
@eddiemonge Ah, you're right, it possibly does involve more configuration when there's a task whose name doesn't in any way map to a plugin (eg. useminPrepare --> grunt-usemin). Although for me this is at least partly balanced out by the fact that jit-grunt makes more attempts at mapping by default, and so with other tasks (eg. assemble) it finds them without any special configuration where load-grunt-tasks does require it.

Speaking as a relative beginner with Grunt / node.js myself, I found it difficult to figure out how to configure load-grunt-tasks to work with a plugin whose name doesn't match load-grunt-task's default mapping. If generator-webapp used jit-grunt you'd have a working example -- commented, even -- in the default Gruntfile that would make it clear how to add new plugins / tasks. Something like:

``` javascript
// Auto-load grunt tasks as they're needed:
require('jit-grunt')(grunt, {
  // Tasks whose names match the name of an installed plugin 
  // (eg. 'task', 'grunt-task', 'grunt-contrib-task') 
  // will be found and loaded automatically.
  // 
  // For tasks whose names don't match the name of a plugin
  // tell Grunt which plugin to load here:
  useminPrepare: 'grunt-usemin'
});
```

...admittedly a _little_ more complex as a default, but by being explicit rather than implicit, probably not actually worse (and arguably even perhaps better) for beginners. Anyway, just a thought / suggestion. For me it doesn't feel like there's that much in it in terms of complexity, and jit-grunt's significant speed improvements certainly make working with grunt more pleasant, on my machine anyway.

eddiemonge:
oh thought this was an issue not a pr. if someone opens an issue ill do more exploring on this

PR #279, opened 2014-02-21
Fix Glyphicons path for Sass Bootstrap

silvenon:
I also added some icons to `index.html` so we can instantly see if Glyphicons loaded properly.


sindresorhus:
Thanks :)

PR #285, opened 2014-02-23
Rename 'yeoman' var to 'settings'

creynders:
Since the former makes my eyeballs bleed ;)

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -19,7 +19,7 @@ module.exports = function (grunt) {
     grunt.initConfig({
 
         // Project settings
-        yeoman: {
+        settings: {
             // Configurable paths
             app: 'app',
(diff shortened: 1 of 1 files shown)


sindresorhus:
settings => config

creynders:
I've been using `config` myself up to now, but I'm a bit worried it might be pretty confusing to people. That they assume it refers to the full configuration object and will be doing `<%= config.connect.test.options.port %>` for instance. Since settings isn't used in grunt yet it seemed like a good name to delineate between user defined values and configuration values.

sindresorhus:
But `settings` could also be mistaken as the full settings object in a confused mind. I don't think it's that much of a problem. I'd rather we try `config` for now.

kevva:
I'm fine with either but @sindresorhus word is the law.

PR #290, opened 2014-02-27
Added images folder to assetsDirs - usemin

fluke:
This is in reference to issue #277 and the solution I found to it. yeoman/yeoman#824

PR #291, opened 2014-03-02
Update bootstrap font copy in gruntfile

cebor:
I think this is the way the fonts from bootstrap should be copied.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -354,11 +354,17 @@ module.exports = function (grunt) {
                         '.htaccess',
                         'images/{,*/}*.webp',
                         '{,*/}*.html',
-                        'styles/fonts/{,*/}*.*'<% if (includeBootstrap) { %>,<% if (includeCompass) { %>
-                        'bower_components/bootstrap-sass-official/vendor/assets/fonts/bootstrap/*.*'<% } else { %>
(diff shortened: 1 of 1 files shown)


sindresorhus:
Why?

cebor:
Because the path in bootstrap.css points to '../fonts/*', and the grunt task compiles this css file it into a vendor.css in '/styles', so the fonts must be in '/fonts' and not in '/bower_components/...'

silvenon:
This is indeed true for LESS Bootstrap, but not for Sass Bootstrap. We could configure `$icon-font-path` to look in the same place for consistency.

cebor:
Oh sorry, i assumed that bootstrap-sass also uses relative paths like bootstrap.

silvenon:
Also, because it's so easy to make a mistake with such a huge-ass Gruntfile, you should always run `grunt serve:dist` to check if everything works first. For example, your `Gruntfile` could have a [syntax error](https://github.com/cebor/generator-webapp/blob/b3dbae84ffe255a4937a80ea63516c389c52b144/app/templates/Gruntfile.js#L244-L245) (not your fault), or a [wrong usemin configuration](https://github.com/yeoman/generator-webapp/pull/290#issuecomment-36469129) (also not your fault).

eddiemonge:
fixed by Pull Request #310: Install bower_components in the root directory

PR #296, opened 2014-03-05
Change prompt order, make Sass and Modernizr false by default

addyosmani:
As discussed with @sindresorhus we think it would make sense to avoid developers opting for the kitchen-sink by giving them a lighter set of initial defaults. This PR:
- Moves Bootstrap up to the first prompt option
- Makes Sass and Modernizr false by default

Our plan is to eventually deprecate the 'Sass with Compass' option in favor of grunt-sass as it's significantly faster and we no longer consider Compass to offer as much value as we once did.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -48,17 +48,17 @@ AppGenerator.prototype.askFor = function askFor() {
     name: 'features',
     message: 'What more would you like?',
     choices: [{
-      name: 'Sass with Compass',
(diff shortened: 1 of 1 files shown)


kevva:
+1

sindresorhus:
:+1:

addyosmani:
:ice_cream:

PR #297, opened 2014-03-06
use bowerInstall on `watch` task.

stephenplusplus:
I stumbled upon this idea when integrating `wiredep` with `generator-gulp-webapp`. It kinda makes sense! If you `bower install` while `grunt watch` or `grunt serve`-ing, the `bowerInstall` task will run and update your files for you. Is this a good idea?


addyosmani:
Sgtm.

PR #300, opened 2014-03-07
Changing the way Bootstrap Sass package is ref'd

ghost:
Aliasing "bootstrap-sass" to the github URL
git://github.com/twbs/bootstrap-sass.git is probably not a good idea.
This could create confusion with the bootstrap-sass package in
the official repository which actually points to
git://github.com/jlong/sass-twitter-bootstrap.git and only goes up to
version 3.0.2 currently. This is NOT the official package.

Changed references to "bootstrap-sass-official" to avoid any
potential confusion.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -122,7 +122,7 @@ AppGenerator.prototype.writeIndex = function writeIndex() {
 
   // wire Bootstrap plugins
   if (this.includeBootstrap) {
-    var bs = 'bower_components/bootstrap' + (this.includeCompass ? '-sass/vendor/assets/javascripts/bootstrap/' : '/js/');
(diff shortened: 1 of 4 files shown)


sindresorhus:
Agree. Thanks :)

ghost:
By the way, I noticed that generator-webapp in the official yeoman repository is quite out of date. How often does it get updated? When will it be updated to the very latest version on Github?

addyosmani:
@sindresorhus I suggest we cut a release and land the RequireJS changes in a follow-up. I've had to rewrite the patch a few times based on numerous issues with grunt-usemin and am not yet happy to PR what I have.

PR #301, opened 2014-03-08
`--split-configs` flag

creynders:
I've been mailing with @addyosmani about investigating the possibility of adding a `--split-configs` flag, which allows you to generate the grunt config into several config files.

The most significant change is that I had to extract the entire grunt config into a separate file (`_config.js`) in order to be able to process it one way or the other. 

I wrote a test, but unfortunately I can't run the tests, I get:

``` shell
Uncaught Error: You don't seem to have a generator with the name webapp:app installed.
```

Not just with my branch, but with the master as well. (I've tried re-installing generator-webapp, npm cache clean, etc. etc. All to no avail though)
I did however test it out pretty thoroughly manually and it works well. 

Let me know what you guys think!


silvenon:
Just out of curiosity, did your npm break like mine did? At one point I realised I can't properly install new generators. Try installing a generator you don't have, e.g. `npm install -g generator-angular` and then `yo --help` to see if it's listed. If it's not showing up, I might know a fix.

creynders:
@silvenon No. Everything works fine, except I can't test. However, I just migrated from a computer with OS X 10.6 to a new one with 10.9 AND I've been f'ing around with various forks of generator-webapp, so it might well be a problem on my system. I have no idea where to start looking though.

creynders:
Argh. I really have to get the testing fixed. Travis fails.

app/templates/_config.js
```diff
@@ -0,0 +1,386 @@
...
+                }]
+            },
+            server: '.tmp'
+        },
+
+        // Make sure code styles are up to par and there are no obvious mistakes
+        jshint: {
+            options: {
+                jshintrc: '.jshintrc',
+                reporter: require('jshint-stylish')
```
silvenon: This is the line where the tests fail, it can't find `jshint-stylish` this way because it doesn't know where to look for. Not really sure how to fix it, though, I'm just pointing it out in case you didn't know :D

creynders: Yeah, it's not a problem when using the generator with `--split-configs` only when testing since `jshint-stylish` isn't a dependency of the generator. I could add it to the devDeps, but that's a bad solution IMO. Hmm don't really see an alternative though.

silvenon: (I commented on the wrong place, so I'm just pasting here.) It **is** a problem when using the generator as well, I get the same error when I try to do `yo webapp`.

creynders: Hmmm, yes of course. I'm a bit baffled now that it did run when I was (hastily and manually) testing it.

creynders:
Ok, a new try at  --split-configs. I tried all kinds of approaches but the `require('jshint-stylish')` kept on biting my ass. The only solution I could come up with was to have the grunt config split into several templates and then either assembling them into one grunt file (by default) or processing them to separate js files in case the --split-configs flag is set.

Just to be clear: this needs MAJOR cleaning up, both code and git history wise. This is just a POC, since I'd rather know if you @addyosmani @sindresorhus  would consider having the config already split up or not. If the latter, no use in pursuing this approach. 

@silvenon Could you take a look if this works for you too ? :) You'll get some jshint errors when running grunt scaffolded w/o the split-configs.

creynders:
And travis passed ! Woohoow ;)

silvenon:
It works, good job :+1: 

There has been a long discussion about splitting tasks into separate files, which you were a part of, so you know that the team wasn't super excited about it, but I'm sure they wouldn't mind having it as a flag.

I would like to have this feature very much, but the problem is that implementation of this can vary, from your approach to [load-grunt-config](https://github.com/firstandthird/load-grunt-config). For example, I like CoffeeScript, but this would require me to convert each file to CS instead of just one (Gruntfile).

Splitting tasks into separate files seems like a job for a CLI command, probably coming from load-grunt-config.

What do you think?

sindresorhus:
I don't think it makes sense to include this in generator-webapp. I would rather see it prove its use-case in the wild as a separate generator first.

creynders:
@silvenon I was planning on including coffee output too. BTW, you do know my [grunt-generate-configs](https://github.com/creynders/grunt-generate-configs) generates coffee files, right? No more manual conversions :)

@sindresorhus I don't think a separate generator will take off TBH. It requires separate promotion, something which I'm not really good at :grimacing:

addyosmani:
Some thoughts:

Although the team aren't yet sold on the split-gruntfile config approach, I personally wouldn't mind this living behind a flag for now. We recently landed support in core for composable generators and `generator-webapp` is going to require a major rewrite in the next two months regardless. That's going to require changing how we handle anything passed as a flag anyway.

If it did land, we would 1. label it as an experimental and 2. need some level of commitment regarding bugs specific to the flag being addressed (team focus is on the long-tail features not behind flags, which I hope makes sense), otherwise it would end up being dropped (like we originally did for RequireJS). Do those terms sound reasonable? 

cc @passy @stephenplusplus @eddiemonge for their thoughts on this too. It's still up to the team to weigh in on whether they're okay with this being behind a flag for the time being.

creynders:
> Do those terms sound reasonable?

@addyosmani certainly!

If the rest of the team can get behind it, I'll start making this merge-ready. As I wrote above the only extra functionality I'd add is when `--coffee --split-configs` is passed I'd let it generate `.coffee` files instead of `.js`. Can you guys live with the added `js2coffee` dep?

creynders:
I used `js-beautify` to fix the indentation problem of the configs (3 tabs needed when in the grunt file, but only 2 in the separate .js files)  but I was wondering why the entire grunt file isn't passed through beautify? Less formatting problems caused by the underscore template variables.

This is nearing completion BTW. I just have to update the docs and make sure the latest changes are merged into this.

creynders:
@passy @stephenplusplus @eddiemonge What do you guys think?

app/index.js
```diff
@@ -4,18 +4,43 @@ var path = require('path');
 var spawn = require('child_process').spawn;
 var yeoman = require('yeoman-generator');
 var chalk = require('chalk');
-
+var js2coffee = require('js2coffee');
+var beautify = require('js-beautify' ).js_beautify;
```
corburn: There is an extra space in the require statement.

silvenon:
Maybe you or I could sum it up, it will seem more stable and attractive.

silvenon:
And if it's done, you could remove "WIP" from the PR's title :sunglasses:

addyosmani:
@sindresorhus @passy Let's talk about this one at the next team meeting and figure out how to proceed.

@creynders thanks for your patience (btw, could you squish and rebase your commits please)? :)

creynders:
@corburn Thanks for the catch!

@silvenon unWIPped :)

@addyosmani NP. The PR's squashed, rebased and cleaned up. I still can't run the tests though. And, for some reason, travis doesn't seem to pick this up for testing either.

creynders:
Ah, never mind the comment about travis, it was picked up after all.

creynders:
@addyosmani if you guys deem it's a problem the grunt template is split into several config templates, I could investigate the possibility of leaving everything as-is and only after the gruntfile is generated through the generator-webapp use esprima with escodegen or a modified version of ast-query to replace the grunt configuration object in the generated grunt file with a `require('load-grunt-configs')` call. I'm not sure whether that would work, but it might be an interesting different approach.

creynders:
@addyosmani Have you guys had the time to look into this?

addyosmani:
@creynders We've had time to review the implementation. My current position is that I personally don't mind this being available optionally but other team members have concerns about the long-term ease of maintainability and being able to make changes once it lands (which is legit). I would like to hear +1 or -1s from the team on whether this should land at this time as I know you've been waiting patiently.

cc @stephenplusplus @passy @SBoudrias @eddiemonge @sindresorhus

sindresorhus:
:-1:

stephenplusplus:
We run into these situations where we'd like to give developers what they want, (RequireJS, CoffeeScript, Jade, Sass, Stylus, Gulp, Grunt, etc.) but have to resist because of what it means for the maintenance and development of the generator, and the general scope of what a generator should provide. For those who prefer splitting their Grunt config into multiple files, they can still integrate it with the same effort as if they weren't using Yeoman. Developers are meant to use the scaffolded boilerplate as their jumping off point (who wants to stand around on a boiling-hot plate?), and take their app in any direction. Yeoman still gives them a head-start, whether they choose to split or not, which to me means we're doing what we're supposed to. Thanks for all of the effort you've put into this and your patience. But from me, I have to give a :-1:

eddiemonge:
:+1: for the idea but :-1: for this implementation. Having https://github.com/yeoman/generator-webapp/pull/301/files#diff-8b50ebbf3a510307ebab0715e7d7aec4R29 still feels like a full config.

Plus, how would this play the new Gruntfile API?

creynders:
Haha, ok that's pretty clear.

@sindresorhus care to explain a little more? Are you against the whole idea or the implementation?

@stephenplusplus If I'd manage to pull this off with as little change to the codebase as possible, would you change your mind or do you think it's just a bad idea?

@eddiemonge I don't understand the full config comment. You mean for you, the Yeoman-team, it still feels like having a full config?

eddiemonge:
https://github.com/yeoman/generator-webapp/pull/301/files#diff-8b50ebbf3a510307ebab0715e7d7aec4R29 I'd prefer something that did it automatically. More like how load-grunt-tasks does it.

sindresorhus:
We're gonna pass on this. I'm sure some would appreciate a split up Gruntfile, so I would encourage you to create a generator using it, but not a good fit for this generator.

PR #304, opened 2014-03-13
Fix documentation for bower-install

rzschoch:
Fix documentation for executing grunt bower-install


stephenplusplus:
Good catch. Thanks!

PR #310, opened 2014-03-20
Install bower_components in the root directory

silvenon:
This is build on top of #263.

It seems something was fixed in the meantime (wiredep maybe) because grunt-bower-install is generating correct path (starting with `../bower_components`), so no ugly fixes were necessary.

I was paranoid and tested everything manually (on top of running tests), every path should now be correct :smile:


silvenon:
This should also close #291.

silvenon:
@eddiemonge Was this the proper way to do this PR? It feels weird that there are now two PRs with the same name...

eddiemonge:
awesome. thanks

PR #312, opened 2014-03-20
Downgrade imagemin to a more stable version

silvenon:
grunt-contrib-imagemin 0.5.0 randomly fails by throwing an error or generating empty images. gruntjs/grunt-contrib-imagemin#163


eddiemonge:
Wouldn't a better fix be to fix grunt-contrib-imagemin so it works?

sindresorhus:
@eddiemonge yes, but it's just waiting for a new version of `image-min` which should be a lot more stable and faster. For now, let's just revert.

Issue #302, opened 2014-03-10
Option for Sass without Compass


fluke:
As mentioned in #294 compass runs really slow. One of the suggestions was to use https://github.com/sindresorhus/grunt-sass and I have implemented it. It's much faster now.

Could we have another option for Sass without Compass?

I'm trying to make a custom version for myself. How do I add the option for Sass without Compass to the CLI and set a variable like includeCompass, so as to make the appropriate templates?

silvenon:
That would be a very good improvement. Everything starts from `index.js`, so you can read the docs on [yeoman.io](http://yeoman.io) to see how it works.

The Yeoman team seems to be very careful about how options in this project are structured, so it may be best to wait for their suggestions before submitting a PR.

fluke:
Sure, I'll have a look, if you could spread the idea I would be grateful.

silvenon:
Well, I'm sure the team got the notified, we'll ping them if they don't respond for a while :D

fluke:
Alright. I'll work to make my own generator and I'll submit a pull request if I manage.

addyosmani:
How are things going @kartikluke?

fluke:
This slipped my mind completely, I found it much easier to copy paste the Gruntfile from my other projects. I'll try making it. As silvenon said 

> The Yeoman team seems to be very careful about how options in this project are structured, so it may be best to wait for their suggestions before submitting a PR.

I decided to to wait as he advised.

silvenon:
Yeah, this one is a bit tricky because Compass includes Sass, so on `yo webapp` we can have:
- Sass
- Sass with Compass
- Bootstrap
- Modernizr

But then you can select both "Sass" and "Sass with Compass", which is weird. Maybe we could deselect "Sass" if "Sass with Compass" is selected.

Or perhaps:
- Sass
- Compass
- Bootstrap
- Modernizr

Where "Sass" gets selected automatically (if not already selected) when "Compass" is selected. This is a less weird solution, in my opinion.

fluke:
I'm working on building this. As for the options. I've made it as follows: You can select Sass if Sass is selected it gives you the option to add Compass support as well. The inquirer.js code I've used is below.

```
var prompts = [{
    type: 'checkbox',
    name: 'features',
    message: 'What more would you like?',
    choices: [{
      name: 'Bootstrap',
      value: 'includeBootstrap',
      checked: true
    },{
      name: 'Sass',
[...]

PR #315, opened 2014-03-21
Add option for Sass without Compass

fluke:
I've implemented it such that you choose Sass. And then you're shown a confirm prompt asking if you want Compass. I've let the majority of the code remain the same in the templates the new Sass code is demarcated by `includeSass && !includeCompass`.

**Side by side comparisons**

Execution Time (2014-03-21 08:48:27 UTC)
    sass:server    342ms  98%
    Total 349ms

Execution Time (2014-03-21 08:47:20 UTC)
    compass:server  20.4s  100%
    Total 20.4s

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -52,26 +52,37 @@ AppGenerator.prototype.askFor = function askFor() {
       value: 'includeBootstrap',
       checked: true
     },{
-      name: 'Sass with Compass',
-      value: 'includeCompass',
+      name: 'Sass',
+      value: 'includeSass',
       checked: false
     },{
       name: 'Modernizr',
       value: 'includeModernizr',
       checked: false
     }]
+  }, {
+    when: function (answers) {
+      return answers.features.indexOf('includeSass') !== -1;
+    },
+    type: 'confirm',
+    name: 'libsass',
+    value: 'includeLibSass',
+    message: 'Would you like to use libsass? Read up more at \n' + chalk.green('https://github.com/yeoman/generator-webapp/blob/master/libsass.md'),
+    default: false
   }];
 
   this.prompt(prompts, function (answers) {
     var features = answers.features;
 
     function hasFeature(feat) { return features.indexOf(feat) !== -1; }
-
     // manually deal with the response, get back and store the results.
     // we change a bit this way of doing to automatically do this in the self.prompt() method.
-    this.includeCompass = hasFeature('includeCompass');
+    this.includeSass = hasFeature('includeSass');
     this.includeBootstrap = hasFeature('includeBootstrap');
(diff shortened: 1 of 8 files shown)


fluke:
Fixed the tests.

silvenon:
I tried it, the Compass option doesn't appear for me when I select Sass. Also, it may not be a good idea to hide options from users, because they won't know Compass is available until they select Sass.

eddiemonge:
see https://github.com/yeoman/generator-webapp/issues/302#issuecomment-38306177

fluke:
Getting an odd issue with Inquirer.js Moving through the options for the follow up question according to https://github.com/yeoman/generator-webapp/issues/302#issuecomment-38351275 causes a new prompt to be made with the cursor moved in whatever direction we wanted. https://github.com/SBoudrias/Inquirer.js/issues/102

silvenon:
Very cool. I noticed a couple of things:
- Why did you selected Modernizr and Sass by default? @addyosmani explicitly unselected them. (For some reason :P)
- libsass is C, not Node.js
- I would capitalize Ruby, because that's the official name of the language (this is obviously a very important change :P)

Also, very cool that you noticed that grunt-sass and grunt-contrib-sass config as slightly different :+1:

addyosmani:
For historic info: we unselected the Modernizr and Sass options by default as found that most users only want the thinnest base possible and it's mostly power-users that want these :)

fluke:
Oh I did that so I wouldn't have to select Sass to see if the follow up question would show correctly. Simply convenience for me :P. I'll change that.

> This task uses the experimental and superfast Node.js based Sass compiler node-sass (which only compiles .scss files).

I got confused by that.

addyosmani:
Can we rebase this and squash the commits so it's easier to merge?

fluke:
Squashed and rebased.

addyosmani:
Still gray on this end. Sure it's rebased?

silvenon:
@addyosmani Are you down with the [final solution](https://github.com/yeoman/generator-webapp/issues/302#issuecomment-38428481)? Because that's the one that got implemented, I just want to make sure you agree before this gets merged.

addyosmani:
I'm down with the final solution. cc'ing @sindresorhus for a final sanity check.

fluke:
Whoops only squashed it. Rebased it. @addyosmani

addyosmani:
:star: :star: :star: 

Thanks for all the work getting this in everyone!

sindresorhus:
Thanks @kartikluke :)

fluke:
Thanks everybody. Always wanted to contribute something substantial to open source.

fluke:
@cjerdonek This is the PR. You can see the commits related to this. #302

PR #323, opened 2014-03-29
templates _package.json updated to use latest version of all packages

lhwparis:
required node version bumped to 0.10 because grunt-contrib-imagemin requires it.


addyosmani:
I'm happy with us bumping to 0.10 but we should make sure to add a note to the release notes clearly stating this so it doesn't trip folks up. Thanks for the PR @lhwparis!

PR #333, opened 2014-04-07
Don't try to copy bowerrc, it causes a fatal error

silvenon:
Because it doesn't exist anymore.


sindresorhus:
oops, yeah, forgot that. thanks :)

PR #335, opened 2014-04-07
Clean up bowerInstall

silvenon:
The `ignorePath` in the `app` target is unnecessary because `bower_components` is no longer in the `app` folder, so it doesn't do anything anymore.

The `ignorePath` in `sass` target doesn't do anything either, because the generated path starts with `../../bower_components`, like in the diff.

Both of these work without `ignorePath`, I tested both `serve` and `build`.

I believe I got it right this time :wink:


addyosmani:
LGTM. Thanks @silvenon!

PR #342, opened 2014-04-15
Make the bootstrap font path relative

richardwestenra:
The current Bootstrap font path only works for projects in root directories. I recommend making the path relative instead of absolute so that it will also work when projects are deployed to subdirectories.


silvenon:
:+1:

stephenplusplus:
@richardwestenra, your review of #343 would be most welcome :+1:

richardwestenra:
@stephenplusplus Sure. Both the "../bower_components" path (as used on #342) and the "../../bower_components" path (as used on #343) work equally well when serving the app locally. However it only works on a dist build when you use the single ../.

It'd be worth a look from someone more experienced with Yeoman: I'm still relatively new to it.

PR #343, opened 2014-04-15
closes #342 - remove loadPath & includePath.

stephenplusplus:
We recently removed `loadPath` from the Gulp generator to avoid unnecessary magic and confusion. Should we kill it here, too?


kevva:
+1 for consistency between them.

PR #345, opened 2014-04-15
Fix compile sass files with .sass extension

artofhuman:


eddiemonge:
better format would be like whats on https://github.com/artofhuman/generator-webapp/blob/master/app/templates/Gruntfile.js#L42

Another questions is do people use .sass extension?

kevva:
Do people use Sass?

addyosmani:
@kevva They do. We've decided recently to move away from using Compass but Sass is still afaik heavily used.

eddiemonge:
I think the question should have been:

> Do people use `.sass`?

kevva:
No, I was just kidding around, hehehe.

eddiemonge:
I wasn't. My question still stands

eddiemonge:
@artofhuman make the code like this: `['*.{scss,sass}']` and I'll accept it.

Sidenote, why are the `server` and `dist` files for _Sass_ the same?

artofhuman:
@eddiemonge /cc

People use sass ext. =) For example author autoprefixer use sass in https://github.com/ai/evil-front

PR #351, opened 2014-04-30
Fix location for Bootstrap fonts with Sass and CSS

sygint:
The entire bootstrap-sass fonts folder path was being copied across, now it copies just the files, it works with regular bootstrap as well.


addyosmani:
Thanks. Could you please squash your commits before we go ahead with a merge?

PR #354, opened 2014-04-30
Clean up vendor.css build comments

sygint:
Keeps from creating unneeded usemin comments, which in turn keeps grunt from trying to process the file


sygint:
After realizing how this works, I see why it is here, closing.

PR #357, opened 2014-04-30
Add main.js to index.html

sygint:
We bother to setup main.js through index.js, but we never even include it, might as well add it to the foot of the index.html?


eddiemonge:
We do write it, in these lines: https://github.com/yeoman/generator-webapp/blob/master/app/index.js#L163-L169

sygint:
yea, i saw that later, my mistake

PR #367, opened 2014-05-10
bootstrap.scss path is wrong in main.scss

albatrosary:
An error occurred while After running the grunt task.

$ grunt serve

...

Running "concurrent:server" (concurrent) task

```
Running "copy:styles" (copy) task


Done, without errors.


Execution Time (2014-05-10 05:29:26 UTC)
loading tasks  6ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 38%
copy:styles    9ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 56%
Total 16ms
    Warning: Syntax error: File to import not found or unreadable: ../../bower_components/bootstrap-sass-official/vendor/assets/stylesheets/bootstrap.scss.
              Load paths:
                /Users/../Projects/sample
                /Users/../Projects/sample/app/styles
                /Users/../Projects/sample/bower_components
        on line 4 of app/styles/main.scss
  Use --trace for backtrace. Use --force to continue.

    Aborted due to warnings.
```

Execution Time (2014-05-10 05:29:24 UTC)
concurrent:server  1.9s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 99%
Total 1.9s

$

### DIFF
--- a/app/templates/main.scss
+++ b/app/templates/main.scss
@@ -1,7 +1,7 @@
-<% if (includeBootstrap) { %>$icon-font-path: "../../bower_components/bootstrap-sass-official/vendor/assets/fonts/bootstrap/";
+<% if (includeBootstrap) { %>$icon-font-path: "../bower_components/bootstrap-sass-official/vendor/assets/fonts/bootstrap/";
 
 // bower:scss
-@import '../../bower_components/bootstrap-sass-official/vendor/assets/stylesheets/bootstrap.scss';
+@import '../bower_components/bootstrap-sass-official/vendor/assets/stylesheets/bootstrap.scss';
 // endbower
 
 .browsehappy {


addyosmani:
Could you let us know what version of the generator you are using?

magnusbae:
Do you have a .bowerrc file? I'm testing an upgrade from version 0.4.7 to 0.4.9 and the bower_components folder moved. Delete your .bowerrc and it should work as expected. Version 0.4.9 does not generate a .bowerrc file.

albatrosary:
This problem has been resolved After removing the .bowerrc file. 
It was my lack of confirmation. 
Sorry..

WillsB3:
I had this problem too, however there was no .bowerrc file in my project. I'm using the latest master of generator-webapp (currently https://github.com/yeoman/generator-webapp/tree/63b821f49fa1122ea562a6d3436917a46f76a968). 

I fixed this by adding a .bowerrc file to the root of the project, containing the following:

``` JSON
{
    "directory": "bower_components"
}
```

My only explanation is that it seems bower has started installing dependencies in a different location. 
I ran a test and created a new folder and then installed jQuery via Bower. As can be seen from the attached screenshot bower automatically created an `app` directory within my test folder and then installed jQuery in there. I've only just started properly using Bower so not sure if this a new behaviour or something it's been doing for a while. Anyway, it seems this is the problem.

I'm using Bower 1.3.3 and the latest generator-webapp.

screen-shot-2014-05-24-at-15 46 00

magnusbae:
what is the output of `ls -la` in your terminal window? Files prefixed with a period is hidden during normal ls or in Finder.

WillsB3:
@magnusbae When checking if my project had a `.bowerrc` already (before I added my own) I was using Forklift with Show Hidden Files turned on, so I can guarantee there was no `.bowerrc` file in there (before I added mine...).

I've literally only just started the project, and did this using yeoman and the HEAD version of generator-webapp, so the _only_ way a `.bowerrc` file would have been in the project (again, before I added my own to fix this issue) would be if a `.bowerrc` file was present in the `generator-webapp` repository...

magnusbae:
Sorry for asking then. When I updated from 0.4.7 to 0.4.9 (using npm install generator-webapp) I created a new template and moved over the generated files to my existing project. The generated project did not contain a bowerrc-file and installed bower-files to the project-root bower_components folder. That being said - your project folder looks suspiciously empty (edit: re-read your post. Technically this would be a bower issue, but read on). Bower looks for higher levels .bowerrc-files, and executes their paramaters. Go hunting and your problem should be resolved. Basically this is not an issue with the generator, but (as far as I can tell) expected behaviour on bowers part. 

image

WillsB3:
@magnusbae I was just trying to be clear in my explanation, I wasn't having a go at you for asking. Apologies if my post came across otherwise. 

The screenshot showing finder above is not showing my project, just a test folder that I used to test (what I thought was) the default behaviour of bower. My project folder contains all the things that yeoman and `generator-webapp` puts there, and this is not shown in my screenshot above.

Thank you for the hint about bower checking for a `.bowerrc` file in parent directories, I wasn't aware of this! Stupidly I inadvertently had a `.bowerrc` on my desktop which I didn't see before so that's why it was installing stuff in an `app` directory. I also started my project form a directory on the desktop so it also suffered the same problem as it too read the `.bowerrc` file from the Desktop above it. 

Now I remember, I think I checked out ng-boilerplate to my desktop to other day and I'm fairly certain that's where the `.bowerrc` file that remained hidden on my desktop came from.

In short.. Make sure there are no `.bowerrc` files in any parent directories of your project :+1:

PR #371, opened 2014-05-12
Bumping grunt-contrib-imagemin version and adding javascript image revving

magnusbae:
First pull request, so bear with me if it should have been two separate pull requests due to the two commits. I could have squashed them, but it seemed more logical to keep them separate.

Imagemin was broken on Windows (at least) in 0.4.9, bumped to version 0.7.0 (from 0.6.0). Tested on OS X, Windows and Heroku (Linux). 

I've also added usemin to JavaScript files so that when one generates html in JS it still works afterwards. There is a pattern match and it should only grab URI's that are legitimate. 

All tests pass, and no new complaints from JSHint.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -266,10 +266,16 @@ module.exports = function (grunt) {
         // Performs rewrites based on rev and the useminPrepare configuration
         usemin: {
             options: {
-                assetsDirs: ['<%%= config.dist %>', '<%%= config.dist %>/images']
+                assetsDirs: ['<%%= config.dist %>', '<%%= config.dist %>/images'],
+                patterns: {
+                    js: [
+                        [/(images\/.*?\.(?:gif|jpeg|jpg|png|webp))/gm, 'Update the JS to reference our revved images']
+                    ]
+                }
             },
             html: ['<%%= config.dist %>/{,*/}*.html'],
-            css: ['<%%= config.dist %>/styles/{,*/}*.css']
+            css: ['<%%= config.dist %>/styles/{,*/}*.css'],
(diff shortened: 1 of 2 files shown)


sindresorhus:
Congrats on your first PR :dancer: 

It's generally preferred to have unrelated changes in separate PRs, but no worries.

Would you be able to add a test? I imagine replacing assets in JS can be a bit brittle, not sure if it's something we should do by default?

// @SBoudrias @silvenon

magnusbae:
As for the tests I don't have the luxury of time at the moment - had to make the change for the sake of my time in the first place. I think the biggest issue to test for might be cornercases, but it works for me at least. 

The imagemin bump is by far the most important of these two, but having the ability to use images directly from JavaScript is a quite common use case in certain projects. Maybe it could be a toggle during setup?

eddiemonge:
I think doing the regex on css would be better and then maybe have your JS add/remove classes that have the images?

magnusbae:
@eddiemonge Not sure I'm following you here with regards to the regex-part. Usemin is functional on CSS classes. My use case at least was that I want to be able to generate HTML-content in JavaScript, particularily inserting img-tags.

cgcgbcbc:
This pull request has been proposed for one week...It seems that Imagemin's problem would have affected 3 085 people according to [npm](https://www.npmjs.org/package/generator-webapp)..So how about separate the second commit to another pull request and merge that soon?

sindresorhus:
> As for the tests I don't have the luxury of time at the moment

We would be happy to look at this again, but tests are required for us to easily confirm your changes actually work.

Did the other change here: https://github.com/yeoman/generator-webapp/commit/63b821f49fa1122ea562a6d3436917a46f76a968

PR #373, opened 2014-05-14
add --allow-remote option to serve task

cgcgbcbc:
Now users can run `grunt serve --allow-remote` to start a remote accessible server.
I've test it manually, but I've no idea how to modify test.js to cover this...

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -414,7 +414,7 @@ module.exports = function (grunt) {
     });
 
 
-    grunt.registerTask('serve', function (target) {
+    grunt.registerTask('serve', 'start the server and preview your app, --allow-remote for remote access', function (target) {
         if (grunt.option('allow-remote')) {
             grunt.config.set('connect.options.hostname', '0.0.0.0');
         }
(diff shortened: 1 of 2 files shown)


sindresorhus:
In what situation would you need this to be dynamical? Can't you just edit the port in the gruntfile?

cgcgbcbc:
> In what situation would you need this to be dynamical? Can't you just edit the port in the gruntfile?

First of all, some front end designer want to show others the work before push his change to the repo. It may help if he has little knowledge about javascript or grunt.

Secondly, for security concern, I think it's not a good idea if someone change the host to 0.0.0.0 and commit the change into VCS. Then he should edit the Gruntfile when he need to enable remote access and then edit it back.

sindresorhus:
Good point. Mind updating the readme too?

cgcgbcbc:
> Good point. Mind updating the readme too?

I'm considering about it... But I can't decide where to update...
This line `Run grunt for building and grunt serve for preview*` or in `grunt-serve-note`? Can you give me some advice?

sindresorhus:
The former.

cgcgbcbc:
> The former.

updated.

sindresorhus:
Thanks :)

cgcgbcbc:
> Thanks :)

It's my honor :smile:

PR #379, opened 2014-05-19
Usemin for all .html files, not just index.html

sygint:
Why just index.html?


sindresorhus:
Because with front-end web-apps, which this generator targets, you usually only have one .html file. The rest happens in the client. Users are free to change it themselves though.

PR #382, opened 2014-05-29
Add support for custom directories

sygint:
This branch adds support for custom directories via the flag "--config=[file]", the format for this file is:
{"assets":"assets","styles":"css","scripts":"js","images":"img","vendor":"vendor"}

Along with this branch comes a repository for a generator that generates these config files for you via yeoman:
https://github.com/gnarmedia/generator-webapp-config

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -30,9 +30,7 @@ var AppGenerator = module.exports = function Appgenerator(args, options) {
 
   this.options = options;
 
-  this.pkg = JSON.parse(this.readFileAsString(
-    path.join(__dirname, '../package.json')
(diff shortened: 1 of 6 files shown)


eddiemonge:
This can already and should be handled by generator.config which uses Yeoman [configstore](https://github.com/yeoman/configstore)

sygint:
I have rewritten it to use configstore, now it uses just '--config' and finds the file via configstore.  I am in the process of rewriting the custom-directories branch to do the same https://github.com/gnarmedia/generator-webapp-config

PR #386, opened 2014-06-04
Fix components relative path

neilhem:
Fix bower components relative path

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -41,7 +41,7 @@ This works if the package author has followed the [Bower spec](https://github.co
 
 To manually add dependencies, `bower install --save depName` to get the files, then add a `script` or `style` tag to your `index.html` or another appropriate place.
 
-The components are installed in the root of the project at `/bower_components`. To reference them from index.html, use `src="bower_components"` or `src="/bower_components"`. Treat the `bower_components` directory as if it was a sibling to `index.html`.
+The components are installed in the root of the project at `/bower_components`. To reference them from index.html, use `src="../bower_components"` or `src="/bower_components"`. Treat the `bower_components` directory as if it was a sibling to the parent of `index.html`.
 
 *Testing Note*: a project checked into source control and later checked out needs to have `bower install` run from the `test` folder as well as from the project root.
 


silvenon:
Thanks! Could you also change the last sentence of that paragraph to resemble the change? Maybe to this:

> Treat the `bower_components` directory as if it was a sibling to the parent of `index.html`.

You can use `git commit --amend` to add that change to this same commit.

silvenon:
Awesome. There are 3 commits now, can you please [squash](http://davidwalsh.name/squash-commits-git) them?

neilhem:
@silvenon deleted unnecessary commits

silvenon:
Perfect, thanks!

eddiemonge:
I am rolling this back as it is incorrect. `grunt serve` serves the bc folder as if it was in /app so it is a sibling of `index.html`, not a sibling of the parent

silvenon:
Oh, that's right, but the path should indeed be `../bower_components` because otherwise building won't work.

eddiemonge:
The only place `../bower_components` should be used is in the gruntfile for building, but even then, its `./bower_components` relative to the gruntfile. Ok so maybe it also needs to be `../bower_components` in the `styles` folder to reference a component, but that has always been the case, nothing has changed there. `index.html` is still a sibling of the folder.

silvenon:
connect is serving `bower_components` as a sibling to `index.html`, but when building, usemin is looking at the directory structure, not how connect is serving it. Using `bower_components` instead of `../bower_components` in `index.html` will break the build process. Fortunately, connect manages to serve `../bower_components` as `./bower_components`.

This project is using `../bower_components` in `index.html` for that reason.

eddiemonge:
Usemin needs to be updated to use the style, if it doesnt already. Paths shouldnt have to change specifically for the build tool. The build tool should be smart enough (or configurable enough) to determine where the files live on the file system, while the server serves them from the web server path

silvenon:
Well AFAIK it doesn't currently :smiley: I agree both usemin and useref should build following `connect`ed paths, that would put an end to this eternal confusion.

PR #401, opened 2014-06-27
"../"-regex also matched the ts/ of bower_components/ - therefore removing it

djfarly:
The `ignorePath` in the wiredep configuration matched the ending of the bower_components/ folder, removing the "ts/":

``` html
<script src="bower_componenjquery/dist/jquery.js"></script>
```

which obviously should be:

``` html
<script src="bower_components/jquery/dist/jquery.js"></script>
```

discussed here: stephenplusplus/grunt-wiredep#87 - thanks to @stephenplusplus for pointing out :grin:

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -236,7 +236,7 @@ module.exports = function (grunt) {
     // Automatically inject Bower components into the HTML file
     wiredep: {
       app: {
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thanks @djfarly!

eddiemonge:
oops. yeah ../ means any two characters followed by a slash. my bad

PR #405, opened 2014-07-02
Move bower.json logic to app/index

eddiemonge:
Update Bootstrap to 3.2.0
Remove jQuery since bower installs it as a dep of Bootstrap
Move `bower.json` code into the app and create the file from there. The BS plugin list is only needed for the non-Sass version since the Sass version has them listed in the main section of package.json


sindresorhus:
:+1:

app/templates/index.html
```diff
@@ -79,7 +79,6 @@ <h3 class="text-muted"><%= appname %></h3>
 
     <!-- build:js(.) scripts/vendor.js -->
     <!-- bower:js -->
-    <script src="bower_components/jquery/dist/jquery.js"></script>
     <!-- endbower -->
```
binwilly, line 82: This could be removed  `<!-- bower:js -->`

eddiemonge, line 82: thats still needed for grunt-wiredep

PR #411, opened 2014-07-17
exclude 'vendor' folder from bootstrap assets path

neilhem:
As is in bootstrap 3.2 sass version there is no more vendor folder inside assets


silvenon:
Thanks, totally forgot about that! :smiley:

PR #422, opened 2014-08-05
Fetch `.htaccess` file via `npm` and remove the `404` page

alrra:


eddiemonge:
:+1:

app/templates/Gruntfile.js
```diff
@@ -361,11 +361,13 @@ module.exports = function (grunt) {
           dest: '<%%= config.dist %>',
           src: [
             '*.{ico,png,txt}',
-            '.htaccess',
```
sindresorhus, line 364:
```
'../node_modules/apache-server-configs/dist/.htaccess'
```

?

app/templates/Gruntfile.js
```diff
@@ -361,11 +361,13 @@ module.exports = function (grunt) {
...
             'images/{,*/}*.webp',
             '{,*/}*.html',
             'styles/fonts/{,*/}*.*'
           ]
+        }, {
+          dest: '<%%= config.dist %>/.htaccess',
+          src: 'node_modules/apache-server-configs/dist/.htaccess'
```
sindresorhus: Too late to change now, but why put it in a `dist` folder? This only makes the long path even longer without any benefit that I can see.

alrra: > Too late to change now

@sindresorhus That is kinda the idea, to hopefully not need to change it in the future.

> but why put it in a dist folder? 

@sindresorhus In terms of repository organization, I'm not a big fan of mixing the source / distribution files with files that are usually kept in the root (e.g.:  `CHANGELOG.md`, `.travis.yml`, etc.). Plus, at one point I want to automate some of the stuff and provide different distributions, so I thought I better just go with `dist` (like for example [jQuery](https://github.com/jquery/jquery/blob/269a27c70204c7d233eac3cd91a383e9b5759a2f/package.json#L6)  did), as otherwise, if I continue to keep them in the root, it would just become a mess at one point.

app/templates/Gruntfile.js
```diff
@@ -351,6 +351,23 @@ module.exports = function (grunt) {
...
+        },
+        options: {
+          replacements: [{
+            pattern: /# ErrorDocument/g,
+            replacement: 'ErrorDocument'
+          }]
+        }
+      }
+    },
+
```
eddiemonge: this introduces yet another plugin making the gruntfile larger and more complicated. i dont think i like it.

sindresorhus: Agreed. I don't think this is worth a separate config.

alrra: @eddiemonge @sindresorhus let me know [how you want things](https://github.com/yeoman/generator-gulp-webapp/pull/181#issuecomment-51519989), and I'll update the commit accordingly. :)

eddiemonge: just replace the original line 364 with `'../node_modules/apache-server-configs/dist/.htaccess'`

Although Im not sure what to do about the 404 line. I kind of want to take the .htaccess file out completely

sindresorhus: I think we should just drop the 404 page. People shouldn't use it anyways. 

@addyosmani @passy @stephenplusplus @silvenon @SBoudrias you ok with that?

silvenon: Absolutely, drop the 404 page.

app/templates/Gruntfile.js
alrra, line 364: > '../node_modules/apache-server-configs/dist/.htaccess'
> 
> ?

@sindresorhus that [won't work](https://cloud.githubusercontent.com/assets/1223565/3867777/893273f6-2010-11e4-9708-f8976606412a.png).

sindresorhus:
Thanks @alrra :)

PR #433, opened 2014-08-31
Fix Sass sourcemap warning

silvenon:
The sourcemap option defaults to `true`.


addyosmani:
:+1: Thank you!

silvenon:
Just to correct myself, `sourcemap` is not a boolean anymore, it has [4 possible values](https://github.com/gruntjs/grunt-contrib-sass#sourcemap). Please check it out just in case `auto` isn't what you want.

PR #441, opened 2014-09-03
'grunt build' now properly writes bower components to vendor files

seyDoggy:
The usemin blocks for vendor.css and vendor.js were pointed at '(.)' instead of '(app)', which resulted in bower components not being written
to these files.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -161,7 +161,7 @@ module.exports = yeoman.generators.Base.extend({
           bs + 'collapse.js',
           bs + 'tab.js'
         ],
-        searchPath: '.'
+        searchPath: 'app'
       });
     }
 
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -7,7 +7,7 @@
     <meta name="viewport" content="width=device-width">
     <link rel="shortcut icon" href="/favicon.ico">
     <!-- Place favicon.ico and apple-touch-icon.png in the root directory -->
-    <!-- build:css(.) styles/vendor.css -->
+    <!-- build:css(app) styles/vendor.css -->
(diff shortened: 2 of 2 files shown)


addyosmani:
I'm surprised that we haven't run into raised issues with bower components not being written correctly to the appropriate vendor files before now. I would have assumed `(.)` in this case would reference app/index.html (and any directories referenced would work against the same path).

@silvenon have you run into issues like this with the current setup?

silvenon:
I'm also surprised that this hasn't come up before, but @seyDoggy is right. Though I wouldn't want to merge this just yet before we figure out what is going on.

If someone can explain to me why Modernizr builds successfully, as it doesn't have a search path defined, I'll be very happy.

eddiemonge:
I dont think search path should be `app`. I'm not sure what is going on without testing this but that seems wrong to me.

seyDoggy:
Not that other generators ought to be set as the bench mark, but there are other generators point to (app) as well.

eddiemonge:
Yeah did another test and it works without this change. What version of usemin do you have in your package.json?

seyDoggy:
Well this is embarrassing. So much time has passed that I can't for the life of me figure out what I was ever using the webapp generator for, so I can't go back and look at the dependency versions. :(

I just updated everything and created a new `yo webapp` and everything builds fine. So something, somewhere, somehow has resolved this issue for me. Still wish I could remember what I was building at the time.

kevva:
So I guess this is fixed then? Happy to hear it works for you now :).

seyDoggy:
I'd say it is, yes.

PR #443, opened 2014-09-10
Added --on option to grunt serve to support selection of browser on star...

IVisualDev:
Hi

i've added a useful option to "grunt serve" to select the browser to start with.
eg. grunt serve --on iexplore

could be useful while developing web apps..

cheers,
Sandro


sindresorhus:
Thanks, but Gruntfile is huge and we're very restrictive on adding new things. This doesn't give enough benefit as most devs mainly develop against their main browser. And you can easily open others yourself.

PR #468, opened 2014-10-26
Fixed issue where container was too narrow

ghost:
I felt like the container looked too narrow on my 1920x1200 display,
and found it was at 730px which seemed wrong. I found this code that
was setting the container's max width to 730px. Once I removed this,
the container was able to size to 970px like it's supposed to.

If this is totally wrong, feel free to reject. But I find it weird that when I size my browser up past 768px, I'm watching the styles live in Firefox inspector, and I see this become active:

```
.container {
  width: 970px;
}
```

But this is also active, thus limiting the width:

```
.container {
  max-width: 730px;
}
```

### DIFF
--- a/app/templates/main.css
+++ b/app/templates/main.css
@@ -65,10 +65,6 @@ body {
 
 /* Responsive: Portrait tablets and up */
 @media screen and (min-width: 768px) {
-  .container {
-    max-width: 730px;
-  }
-
   /* Remove the padding we set earlier */
   .header,
   .marketing,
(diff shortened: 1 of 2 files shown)


ghost:
Thinking about it now, maybe this `max-width` value was defined here so we could easily set whatever we wanted the maximum container width to be?

silvenon:
I'm assuming the `width` property is coming from Bootstrap. They're not using `max-width` at all because it's buggy on older IEs, and the better way would be to configure their container variables. If I had to guess, it's more narrow to improve readability, because the text is going full width.

But in my opinion this is overthinking a scaffolding code which will get deleted anyway.

ghost:
Oh interesting. If I remove most of the `main.scss` content (everything below `// endbower`), it makes no difference on my site. Never mind then. ;)

PR #488, opened 2014-12-10
no need rev the favicon.ico

materliu:
In general, once we choose our site's favicon.ico, we won't chang it. Don't you think so, Sindre?

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -261,7 +261,8 @@ module.exports = function (grunt) {
             '<%%= config.dist %>/styles/{,*/}*.css',
             '<%%= config.dist %>/images/{,*/}*.*',
             '<%%= config.dist %>/styles/fonts/{,*/}*.*',
-            '<%%= config.dist %>/*.{ico,png}'
+            '<%%= config.dist %>/*.{ico,png}',
+            '!<%%= config.dist %>/favicon.ico'
           ]
         }
       }
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -5,7 +5,6 @@
     <title><%= appname %></title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width">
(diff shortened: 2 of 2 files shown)


silvenon:
I change my favicons and they are PITA to update, so I would like to rev it.

materliu:
@silvenon yeah, may be rev is the perfect solution for favicons update condition, but if I forget to use `<link rel="shortcut icon" href="/favicon.ico">` in my html files (assume we have a very large site, and the site has so many html files), things will go wrong.

silvenon:
That's true. I guess the best way to bust favicon cache is using a parameter, like `?v=2`. So it would be better not to rev it then. We'll wait for opinions from others.

materliu:
so thanks for your respond, silvenon.

arthurvr:
Just a little addition to this PR: If we choose to not rev it anymore, [this link](https://github.com/yeoman/generator-webapp/blob/master/app/templates/index.html#L8) isn't anymore needed. If the file is called `favicon.ico`, the browser will automatically find it.

materliu:
@arthurvr  yeah, I've done what you told me.

eddiemonge:
I think rev'ing it is better. @materliu are you creating all your pages manually or using a template? If you have a large site you should be using a template.

:-1:

materliu:
@eddiemonge, yeah, you are right, if we have a large site, we should use a template, but don't forget not only the browser visit our sites, some tools depends on the favicon.ico heavily, for example [1password](https://agilebits.com/onepassword) or [Airmail](http://airmailapp.com/) use it to show the site thumbnail.

eddiemonge:
If those tools don't check the root index file for the favicon then they are broken/doing it wrong.

> The standard implementation uses a link element with a rel attribute in the <head> section of the document to specify the file format and file name and location. Unlike in the prior scheme, the file can be in any Web site directory and have any image file format

eddiemonge:
Closing this since I think the benefits (if any) are outweighed by the cons.

PR #490, opened 2014-12-15
Updated node-sass

sbehrends:
Fixes errors when compiling with libsass

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -33,7 +33,7 @@ module.exports = yeoman.generators.Base.extend({
     if (!this.options['skip-welcome-message']) {
       this.log(require('yosay')());
       this.log(chalk.magenta(
-        'Out of the box you have HTML5 Boilerplate, jQuery, and a ' +
+        'Out of the box I include HTML5 Boilerplate, jQuery, and a ' +
         'Gruntfile.js to build your app.'
       ));
     }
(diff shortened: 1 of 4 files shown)


silvenon:
Could you please squash your commits?

app/index.js
```diff
@@ -45,15 +45,23 @@ module.exports = yeoman.generators.Base.extend({
...
         name: 'Bootstrap',
         value: 'includeBootstrap',
+        checked: false
+      },{
+        name: 'Foundation',
+        value: 'includeFoundation',
+        checked: false
+      },{
+        name: 'Bourbon',
+        value: 'includeBourbon',
```
silvenon, line 55: Why are you adding options for Foundation and Bourbon?

app/index.js
```diff
@@ -33,7 +33,7 @@ module.exports = yeoman.generators.Base.extend({
     if (!this.options['skip-welcome-message']) {
       this.log(require('yosay')());
       this.log(chalk.magenta(
-        'Out of the box I include HTML5 Boilerplate, jQuery, and a ' +
+        'Out of the box you have HTML5 Boilerplate, jQuery, and a ' +
```
silvenon, line 36: Why this copy change?

sindresorhus:
Please keep this PR focused to only updating grunt-sass. Also make sure the tests passes.

sbehrends:
Yup Sorry.
Actually I was extending the webapp generator with Bourbon or Foundation. But while building it found that grunt-sass was giving unexpected errors.

The PR would be just grunt-sass update to 0.17.0 that manage errors correctly.

app/index.js
sbehrends, line 36: Just for fun :)
I thought it as a collaborative fork. So the "first person" does not fit. At least would be "we included". So I change it as "you will receive this with this generator" does not matter who is giving it

app/index.js
sbehrends, line 55: To be more flexbile while building. I found webapp generator the most complete (with img/svg/ minification and so on. The idea is to extend it with other frameworks. (Maybe also adding partials?)

silvenon:
Thanks for your help. You can send small PRs like that straight from GitHub, by just clicking on the edit button when viewing a file. (You can try it now if you want.)

PR #493, opened 2014-12-28
Suggest setting removeRedundantAttributes to false [htmlmin:dist]

jessepinuelas:
Removing redundant attributes from input fields removes attributes that could be styled such as type="text", placeholder etc.. I would suggest setting removeRedundantAttributes to false as leaving it set to true might cause undesired results when distributing/running build

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -323,7 +323,7 @@ module.exports = function (grunt) {
           removeCommentsFromCDATA: true,
           removeEmptyAttributes: true,
           removeOptionalTags: true,
-          // setting removeRedundantAttributes to true will impact styles with attribute selectors
+          // true would impact styles with attribute selectors
           removeRedundantAttributes: false,
           useShortDoctype: true
         },
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -323,7 +323,10 @@ module.exports = function (grunt) {
           removeCommentsFromCDATA: true,
           removeEmptyAttributes: true,
           removeOptionalTags: true,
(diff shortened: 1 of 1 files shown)


app/templates/Gruntfile.js
```diff
@@ -323,7 +323,10 @@ module.exports = function (grunt) {
           removeCommentsFromCDATA: true,
           removeEmptyAttributes: true,
           removeOptionalTags: true,
-          removeRedundantAttributes: true,
+          // removing redundant attributes from input fields removes attributes that could be styled
+          // such as type="text", placeholder etc.. I would suggest setting removeRedundantAttributes
+          // to false as leaving it set to true might cause undesired results when distributing/running build
```
sindresorhus: this needs to be made a lot more terse

eddiemonge: @jessepinuelas ^^

jessepinuelas: @sindresorhus @eddiemonge Roger that, will update shortly.

app/templates/Gruntfile.js
```diff
@@ -323,7 +323,8 @@ module.exports = function (grunt) {
           removeCommentsFromCDATA: true,
           removeEmptyAttributes: true,
           removeOptionalTags: true,
-          removeRedundantAttributes: true,
+          // setting removeRedundantAttributes to true will impact styles with attribute selectors
```
sindresorhus: > setting this to true will impact styles with attribute selectors

hemanth: > ^ true would impact styles with attribute selectors.

eddiemonge:
can you squash the commits please?

PR #495, opened 2014-12-31
Use Grunt APi to render the date in »file header«

mischah:
Makes it less ugly :smile:

Does this make sense to you?


mischah:
Okay, after seeing the travis report I do know why you handling the date that way :sweat_smile:

PR #505, opened 2015-02-06
Only apply `no-js` class when needed

arthurvr:
See https://github.com/yeoman/generator-gulp-webapp/commit/3771228b48382fba4435a22b71722a2c1678547c.


silvenon:
Thanks :+1:

PR #510, opened 2015-02-24
Update compass.md

ingozoell:


silvenon:
It should be:

``` diff
-  'sass': {
+  'compass': {
```

Thanks, but I'd rather keep this named `sass`. Others are renamed to `compass` because they actually call Compass tasks. Wiredep will act the same way with or without Compass

PR #515, opened 2015-03-08
Do not trim trailing whitespace

stevemao:


.editorconfig
```diff
@@ -1,4 +1,3 @@
-# editorconfig.org
```
silvenon, line 1: This is a useful link, I would prefer if you leave it :grin:

silvenon:
Hey :smile: Could you elaborate this change? What's the problem with trimming white space in Markdown files?

.editorconfig
stevemao, line 1: This is to be consistent with other repos. I see this line was removed in other repos.

stevemao:
- to be consistent with other repos
- https://github.com/google/web-starter-kit/commit/4b28eac680d9ee0d656b746ab6239dc47d31a469 would be a good example

stevemao:
Basically two trailing space can generate a new line

.editorconfig
silvenon, line 1: Maybe it was never there in other projects, not intentionally removed. The link is present in generator-gulp-webapp, though. Could you leave it for now? I'll think it's a helpful inconsistency :smiley:

silvenon:
Got it. I'll merge after the comment in the header is added back.

.editorconfig
stevemao, line 1: Sure. I'm not really sure about this. @sindresorhus should know more details.

stevemao:
Done

silvenon:
Thanks!

stevemao:
:smile:

PR #526, opened 2015-04-01
Fix error of Gruntfile with coffee option

htanjo:
The `Gruntfile.js` is collapsed when generating a webapp with `--coffee` option.

``` javascript
concurrent: {
  server: [
    'coffee:dist'    // "," is missing
    'copy:styles'
  ],
  test: [
    'coffee',,       // "," is duplicated
    'copy:styles'
  ],
  ...
```

I fixed the template. Thanks.


htanjo:
Hmmm. The Travis build failed but I don't think it was caused by my code.
It's probably potential error about timeout setting.

PR #531, opened 2015-04-23
Add `npm test`

stevemao:


app/templates/_package.json
```diff
@@ -31,5 +31,8 @@
   },
   "engines": {
     "node": ">=0.10.0"
+  },
+  "scripts": {
+    "test": "grunt test"
```
eddiemonge, line 36: without grunt-cli this will fail. why someone would not have grunt-cli I don't know but if we are going to have a grunt command then we should include the binary to run that command

silvenon, line 36: Should we add `grunt-cli` as a dependency and run it from `node_modules`?

stevemao:
Done

silvenon:
Works, thanks! :+1:

PR #533, opened 2015-04-23
make jasmine test framework work

stevemao:
Previously jasmine framework cannot work without more manual configs. For coffeescript the compiled javascripts are sitting in .tmp folder. Unfortunately unlike mocha, jasmine does not read a index.html to load scripts.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -46,7 +46,7 @@ module.exports = function (grunt) {
       },<% } else { %>
       js: {
         files: ['<%%= config.app %>/scripts/{,*/}*.js'],
-        tasks: ['jshint'],
+        tasks: ['jshint']
       },
       jstest: {
         files: ['test/spec/{,*/}*.js'],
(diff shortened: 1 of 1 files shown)


app/templates/Gruntfile.js
```diff
@@ -93,7 +93,7 @@ module.exports = function (grunt) {
           logLevel: 'silent',
           host: 'localhost',
           server: {
-            baseDir: ['.tmp', './test', config.app],
+            baseDir: ['.tmp', './test', config.app, './'],
```
eddiemonge: this could be a dangerous change

stevemao: Why is this dangerous? `'./'` is the last place it will search at if files cannot be found in `'.tmp', './test' and config.app`. However, it might be better to say "if jasmine then search in `'./'` only.

eddiemonge: why would it be looking in `./`? What is it looking for in there?

stevemao: For jasmine it looks for specs, helpers and vendor etc.
It's the base url for jasmine to look for files

eddiemonge: I feel like specifying a base in the jasmine config is a better choice since its jasmine specific

stevemao: :+1:

stevemao:
Done, thanks

silvenon:
Thanks for this! :+1:

PR #537, opened 2015-04-29
Update test.js

ghost:
Assert was declared twice in test file.

Yeoman assert helper seems to be the most important one so I kept it.

Mistake or not ?


kevva:
Thanks :).

PR #539, opened 2015-05-01
Make sourcemap actually works

stevemao:
There are other ways to do it but this is probably the easiest way because it avoids changing the routes.

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -183,6 +183,7 @@ module.exports = function (grunt) {
     sass: {
       options: {
         sourceMap: true,
+        sourceMapContents: true,
         includePaths: ['bower_components'],
         loadPath: 'bower_components'
       },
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you elaborate how it wasn't working and how this fixes it?

stevemao:
When choosing `ruby-sass` the `sourcemap` option is not enabled. When choosing `node-sass` the `sources` url is different from the routes of the server so it cannot find the sources. We should either change the routes of the server or just make the source inline.

stevemao:
This PR makes the `sources` inline because we have bower_components, .tmp and app/style folders so it might be harder to change the server routes.

stevemao:
Since ruby-sass is removed this is actually more simple

stevemao:
60c27352a50bdc0b2bece0461bdf0e3f5b18a0ea

silvenon:
Damn, I spend many hours fixing this only to come to the same fix which was here all along. Sorry, I didn't see it, but I will give you credit in the release notes for v0.6.1. Thanks for solving it!

stevemao:
Lol no worries. There is some slight difference. Your approach also embeds the sass sourcemap, and let autoprefixer pick up the previous sourcemap by setting it to `true`. I was trying to keep it as close to the original workflow as possible.

silvenon:
Released [v0.6.1](https://github.com/yeoman/generator-webapp/releases/tag/v0.6.1).

Yeah, I wanted to mimic generator-gulp-webapp and thought that there might be less potential problems in the future if sourcemap was inline.

stevemao:
:+1:

PR #543, opened 2015-05-17
Add support for Babel

jamiebuilds:


silvenon:
Hey, thanks for the awesome PR :+1: The team decided to completely replace CoffeeScript with Babel, would you care to add that to your PR too?

jamiebuilds:
Will do

jamiebuilds:
This might be pushing it, but I could make the change that Babel is always included. If people do not want to use ES6 features, Babel would just pass the file right through without transforming anything. It would simplify the repo a bit more and encourage people to start using ES6+.

arthurvr:
Nah. I don't really know if all-day users would like that. Generally I don't like it if my build system is cluttered with things I don't use (in case they wouldn't use ES6), so I don't think they will eighter :/

jamiebuilds:
Alright, removed CoffeeScript

app/templates/Gruntfile.js
```diff
@@ -405,17 +405,17 @@ module.exports = function (grunt) {
...
-      server: [<% if (coffee) {  %>
-        'coffee:dist',<% } %><% if (includeSass) { %>
+      server: [<% if (babel) {  %>
+        'babel:dist',<% } %><% if (includeSass) { %>
         'sass:server'<% } else { %>
         'copy:styles'<% } %>
       ],
-      test: [<% if (coffee) { %>
-        'coffee'<% } %><% if (coffee && !includeSass) {  %>,<% } %><% if (!includeSass) { %>
+      test: [<% if (babel) {  %>
```
arthurvr: Why did you add that space?

app/templates/Gruntfile.js
```diff
@@ -405,17 +405,17 @@ module.exports = function (grunt) {
...
+        'babel'<% } %><% if (babel && !includeSass) {  %>,<% } %><% if (!includeSass) { %>
         'copy:styles'<% } %>
       ],
-      dist: [<% if (coffee) { %>
-        'coffee',<% } %><% if (includeSass) { %>
+      dist: [<% if (babel) {  %>
```
arthurvr: Let's remove that extra space :)

stevemao:
Any reason why you guys want to remove CoffeeScript? Does it apply to other generators?

silvenon:
CoffeeScript was created to patch some shortcomings of JavaScript (other than to provide some syntactic sugar). Now that we have an [ES6 transpiler](https://babeljs.io/docs/learn-es6/), there's no real need to support CoffeeScript anymore. We generally advise using plain JS over CoffeeScript because it's native (i.e. new features will be faster), the barrier of entry is lower (e.g. Atom being written in CS [increases the barrier](https://twitter.com/sindresorhus/status/591869731109343233)) and, in my opinion, there is less and less reason for it to exist.

Yes, we'll probably drop it from other generators as well, eventually.

test/test.js
```diff
@@ -80,26 +78,6 @@ describe('Webapp generator', function () {
...
-
-        assert.fileContent([].concat(
-          expectedContent,
-          [['Gruntfile.js', /coffee/]]
-        ));
-
-        done();
-      });
-    });
-
```
arthurvr, line 102: Should probably test the babel option in some way

arthurvr, line 102: ping? I just think that we should at least have a test for the `babel` option.

jamiebuilds, line 102: added

arthurvr:
I like this :+1: @silvenon WDYT? Good to go?

addyosmani:
LGTM for a merge. 

Side: It's pretty great to see the Babel community proactively helping existing tools improve their ES6 support. Thanks a lot @thejameskyle!

addyosmani:
@silvenon In agreement regarding our general dropping of CS. Now might be a good time for us to open issues for any generators with remaining CS support to consider switching those over to ES6 too.

silvenon:
@addyosmani good idea. Done.

silvenon:
I tried it, seems to work fine. Very good job, @thejameskyle, thanks a lot :+1: 

I would even go as far as not making Babel optional, because it's just JavaScript, but this will do perfectly for now.

addyosmani:
:+1:

jamiebuilds:
Thanks guys!

arthurvr:
> Side: It's pretty great to see the Babel community proactively helping existing tools improve their ES6 support

Sure! :+1: Thanks @thejameskyle!

> In agreement regarding our general dropping of CS. Now might be a good time for us to open issues for any generators with remaining CS support to consider switching those over to ES6 too.

If that's the plan maybe we should consider writing about it at some point too? (a blogpost I mean) It's a change that significantly impacts our users and it's all about shiny new technology. There's much interesting stuff going on in the world of ES2015-tooling these days.

jamiebuilds:
You might also want to switch to eslint with [babel-eslint](https://github.com/babel/babel-eslint) as the parser for better ES6/ES7/JSX/etc support

silvenon:
Thanks for the tip! I've just seen it recently, will check it out :+1:

PR #545, opened 2015-05-20
Add .DS_Store to gitignore

miniplus:
.DS_Store wasn't in .gitignore yet, so decided to add it.


arthurvr:
No thanks. This should be in your global `.gitignore`.

miniplus:
You're right.

PR #565, opened 2015-07-04
Add Gulp support

silvenon:
The goal is to merge [generator-gulp-webapp](https://github.com/yeoman/generator-gulp-webapp) into this generator. Benefits:
- Much DRY. There was a lot of repeating code between these 2 generators.
- Faster development, because it's a single project.
- Ability to deprecate Grunt later on.

Changes besides adding Gulp:
- Turn Babel on by default. I didn't know about the `--no-babel` flag, so I added it to the readme.
- Ask about including jQuery when Bootstrap is off.
- Add `.` to Sass include paths instead of `bower_components`, to mimic generator-gulp-webapp.

Tasks:
- [x] add the functionality
- [x] write tests
- [ ] test it manually
- [ ] transfer issues from generator-gulp-webapp
- [ ] label issues as `gulp` or `grunt`
- [ ] transfer recipes from generator-gulp-webapp
- [ ] update the readme
- [ ] add a landing page (based on [this](http://yeoman.github.io/generator-gulp-webapp/) one)
- [ ] deprecate generator-gulp-webapp and point to generator-webapp

I could use reviews and ideas for this one @sindresorhus, @arthurvr, @kevva.


arthurvr:
We should probably have discussed this before working on it, although, yes. I like this idea :+1:

app/index.js
```diff
@@ -63,9 +61,18 @@ module.exports = generators.Base.extend({
     }
 
     var prompts = [{
+      type: 'list',
+      name: 'taskRunner',
+      message: 'Which task runner would you like to use?',
+      choices: [
+        'Gulp',
```
arthurvr: lowercase `g`?

test/eslint.js
```diff
@@ -4,23 +4,59 @@ var helpers = require('yeoman-generator').test;
...
+      assert.fileContent('package.json', 'es6');
+    });
   });
 
-  it('adds basic configuration', function () {
-    assert.fileContent('package.json', 'eslintConfig');
+  describe('with gulp', function () {
+    before(function (done) {
+      helpers.run(path.join(__dirname, '../app'))
+        .inDir(path.join(__dirname, '.tmp'))
```
arthurvr, line 30: Use `.inTmpDir()` instead? https://github.com/yeoman/yeoman.io/issues/486

test/temp/.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
arthurvr, line 1: the `temp` shouldn't be in the repo right?

app/templates/_package.json
```diff
@@ -1,7 +1,15 @@
 {
   "private": true,
+  "engines": {
+    "node": ">=0.12.0"
+  },
   "devDependencies": {
     "autoprefixer-core": "^5.2.1",
+<% if (taskRunner === 'Gulp') { -%>
```
arthurvr, line 8: Why isn't this just after line 48 like the rest of the gulp-specific things?

silvenon, line 8: Alphabetic order.

arthurvr, line 8: Oh, alright.

test/sass.js
```diff
@@ -4,30 +4,66 @@ var helpers = require('yeoman-generator').test;
...
+          ]
+        })
+        .on('end', done);
+    });
 
-  it('uses SCSS', function () {
-    assert.file('app/styles/main.scss');
-    assert.noFile('app/styles/main.css');
-  });
+    it('uses scss', function () {
```
arthurvr, line 20: Why did you change this? `SCSS` is the correct way IIRC. That's how they write it themselves too.

arthurvr:
In fact I don't see a reason to keep Grunt support except that users might be familiar with it.

sindresorhus:
I don't really see the point of this. I'd much rather just rename `generator-gulp-webapp` to `generator-webapp` and deprecate this one.

If we were to go this way though you should preserve the git history for `generator-gulp-webapp` by combining the histories.

arthurvr:
I tend to agree with @sindresorhus. And honestly, I'd rather see us dropping Grunt than maintaining both. The only reason I think people use Grunt these days is because they don't really care and are just already familiar with Grunt.

sindresorhus:
Grunt does still have its use. It's declarative, so easier for non-coder and beginners, but none of us uses it anymore, so I don't think it makes sense for us to maintain it. We could always ask for a maintainer and move it over to that person.

silvenon:
Oh, I thought this was the right way to do it, I should've asked. Could you further explain your suggestion @sindresorhus? I wanted to merge the generators is to make it much easier to mantain the Grunt version (and `yo webapp` is shorter to type than `yo gulp-webapp`), but deprecation is also fine, as long as we do _something_. If we rename generator-gulp-webapp to "generator-webapp", what will this generator be called?

sindresorhus:
Explain what?

I guess we could rename this one to `generator-webapp-banana`.

silvenon:
Ok. I don't know how that affects npm. Do you get some kind of message that a module has been renamed? Maybe `generator-grunt-webapp` is a better name :smiley:

sindresorhus:
> I don't know how that affects npm. Do you get some kind of message that a module has been renamed?

Yes, we can `npm deprecate` a version of a package, and then users get's a warning when installing with our custom message.

> Maybe generator-grunt-webapp is a better name

I like mine, but I guess this works too :p

PR #596, opened 2015-08-28
Update all fonts references

### LINKED ISSUE #594: App fonts folder location not according to gruntfile
When generating the folder structure there is a `fonts` folder created directly under `app`. I placed my local fonts there but they weren't being copied `dist`, so I looked inside `gruntfile.js` and the paths are different:

``` js
// Renames files for browser caching purposes
    filerev: {
      dist: {
        src: [
          '<%= config.dist %>/scripts/{,*/}*.js',
          '<%= config.dist %>/styles/{,*/}*.css',
          '<%= config.dist %>/images/{,*/}*.*',
          '<%= config.dist %>/styles/fonts/{,*/}*.*',
          '<%= config.dist %>/*.{ico,png}'
        ]
      }
    },
```

and 

``` js
    copy: {
      dist: {
        files: [{
          expand: true,
          dot: true,
          cwd: '<%= config.app %>',
          dest: '<%= config.dist %>',
          src: [
            '*.{ico,png,txt}',
            'images/{,*/}*.webp',
            '{,*/}*.html',
            'styles/fonts/{,*/}*.*'
          ]
```

Unless I'm missing something, the generator should create `app/styles/fonts` instead of `app/fonts` or `gruntfile.js` should be modified.

neilhem:
Fixes #594


eddiemonge:
probably easier to just change this line: https://github.com/yeoman/generator-webapp/blob/master/app/index.js#L288

neilhem:
@eddiemonge I think when generator scaffolds app it creates `fonts` directory inside `app` folder, but references in Gruntfile.js expects fonts directory inside `app/styles`.

silvenon:
I guess it works either way. People like to put them in the `styles` directory because fonts are imported from CSS, but it makes no difference.

neilhem:
So when generator scaffolds app it should do it one way create folder in app/fonts & references in Gruntfile.js should be app/fonts too. If I want to tweak it later, I will move fonts into styles & update references in Gruntfile.

silvenon:
Yep, that works for me. I prefer to have it next to `styles`, rather than in it. Because you may have other directories within `styles` which contain actual stylesheets, and having `fonts` be one of them is weird.

Thanks for the PR!

PR #597, opened 2015-08-29
Fix for usemin JS path that ignores Babel transpilation

### LINKED ISSUE #580: Building a ES2015-based app fails
When doing `grunt` or `grunt build`, the `eslint` accuses things like this:

> app/scripts/main.js
>   1:1  error  Unexpected token let

This is "fixed" by changing the folder `eslint` is searching for do the lint, like from `<%= config.app %>/scripts/{,*/}*.js` to `<%= config.tmp %>/scripts/{,*/}*.js`. This is happening obviously because it is searching within the `app` directory, which has the non-compiled ES2015 code and eslint doesn't know much about it, unless you defined `environment: 'es6'`, but both cases leads to errors like this:

> .tmp/scripts/main.js
>     3:13  error  "Backbone" is not defined

Also uglifying is not working as well. It is attempting to do the same as eslint: doing its task from ES2015 code, which has an unknown syntax for it.

Two alternatives I found so far:
1. Uglifying manually;
2. Avoid ES2015—which this generator is proposed to handle.

By the bye, do you have any suggestions?

NOtherDev:
When running `grunt build`, `grunt-usemin` builds a flow for .js files starting with `scripts` directory that contains not transpiled ES6 code. 
The transpilation (`grunt babel` task) puts transpiled files into `.tmp` directory. It happens before the first step in js flow (`concat`) is run.
Thus, the transpilation works fine, but it's result is ignored and not transpiled files are used in `usemin` flow that ends up in target `dist` directory.

The fix changes the working directory where `usemin` starts its flow to the destination directory of Babel transpilation.

Fixes #580.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -91,7 +91,7 @@
     <!-- endbuild -->
     <% } %>
 
-    <!-- build:js scripts/main.js -->
+    <!-- build:js(.tmp) scripts/main.js -->
     <script src="scripts/main.js"></script>
     <!-- endbuild -->
 


silvenon:
Thanks! :+1:

La-toile-cosmique:
Thank you

Issue #603, opened 2015-09-16
Problem with bower components


StfBauer:
Currently I get the error that components stored in the bower_components cannot be found.

```
Warning: Running "sass:server" (sass) task
    >> file to import not found or unreadable: susy/sass/susy
    >> Current dir: /Volumes/Local/code/customer/app/styles/
    >>   Line 8  Column 9  app/styles/main.scss
```

I tried various different ways to import those components but noting worked. 
In css I tried the following thing and this worked fine in older versions too.

```
@import "susy/sass/susy";
```

Any suggestions? Thanx.

silvenon:
``` scss
@import "bower_components/susy/sass/susy";
```

But `grunt wiredep` should have injected it. Did you run it?

gaboesquivel:
make sure sussy is in bower.json deps for it to be injected thru wiredep.  Wiredep injects the file specified in the main property in the component's bower.json file.  In this case "main": "sass/_susy.scss" , see https://github.com/ericam/susy/blob/master/bower.json

https://github.com/taptapship/wiredep#how-it-works

StfBauer:
Thank you for your replies. Somehow it looks like this issue was related to node version 4.1.0. After I downgraded back to v0.12.7 it works fine again.
The only issue I currently have that browser-sync/livereload is not working.
I see a lot of messages in the console that says.

```
Failed to load resource: Could not connect to the server.
```

The url it tries to load is:
http://localhost:9000/browser-sync/socket.io/?EIO=3&transport=polling&t=1442552633077-27

Worked fine before. Any suggestion on that ?

_Update_: After first run I had that problem. Dug deeper into the problem. Changes in the app directory will refresh the browser but not changes will be stored in the .tmp folder. Updates to the sass file will not reload the browser. Changes for example in the index.html will reload the browser.

gaboesquivel:
yeah, it seems that renaming .tmp to tmp (without the dot) fixes that problem, see https://github.com/yeoman/generator-webapp/issues/585

StfBauer:
tried it and yes it works now without an exception!!
Thank you very much for your support.

To keep the tmp files apart from the other files I renamed it to _tmp. An this works pretty well too.

silvenon:
The fix is released in [v1.1.0](https://github.com/yeoman/generator-webapp/releases/tag/v1.1.0).

Issue #579, opened 2015-08-01
CSS not being injected


ankitjainOfficial:
The generator is running fine, only thing is that the CSS is not being injected and have to do a manual refresh. 

Though, css is being generated. 

```
Completed in 2.435s at Sat Aug 01 2015 13:13:32 GMT+0530 (IST) - Waiting...
>> File "app/styles/main.scss" changed.
Running "sass:server" (sass) task

Running "postcss:dist" (postcss) task
>> 1 processed stylesheet created.

Done, without errors.


Execution Time (2015-08-01 07:45:44 UTC)
loading tasks          482ms  ▇▇▇▇▇▇ 15%
loading grunt-sass     130ms  ▇▇ 4%
sass:server            809ms  ▇▇▇▇▇▇▇▇▇▇▇ 26%
loading grunt-postcss  769ms  ▇▇▇▇▇▇▇▇▇▇ 24%
postcss:dist           948ms  ▇▇▇▇▇▇▇▇▇▇▇▇ 30%
Total 3.1s

Completed in 3.999s at Sat Aug 01 2015 13:15:47 GMT+0530 (IST) - Waiting...
```

silvenon:
Which version are you running? Do you have JavaScript enabled?

ankitjainOfficial:
generator-webapp 1.0.1 and yes, javascript is enabled.

tulsadome:
Im having this exact same problem. I spent a few hours trying to edit the gruntfile with no luck and I ended up switching to gulp-webapp which works perfectly with all the same code.

Caesor:
hi @ankitjainOfficial , have you fixed this error?  I have the same problem.

ankitjainOfficial:
No. I don't have much knowledge with grunt. I generally use gulp, but this
is one of the projects where I needed to use this generator.

philwolstenholme:
I'm also seeing this with version 1.0.1, for a project freshly scaffolded on 14/08/2015.

silvenon:
Probably related to BrowserSync/grunt-browser-sync#117.

silvenon:
It seems that grunt-browser-sync [isn't necessary](http://www.shakyshane.com/javascript/nodejs/browser-sync/2014/08/24/browser-sync-plus-grunt/). I'll implement this solution one of these days, in the meantime you guys can try it if you're in an urgent situation.

hersonls:
Hey guys, I had the same problem and what I notice

[...]

PR #604, opened 2015-09-19
add browserSync watchOptions.ignored

### LINKED ISSUE #585: Change js/css file doesn't trigger browserSync to reload the page
Currently, the livereload property of `browserSync` looks like this:

```
  livereload: {
    options: {
      files: [
        '<%= config.app %>/{,*/}*.html',
        '.tmp/styles/{,*/}*.css',
        '<%= config.app %>/images/{,*/}*',
        '.tmp/scripts/{,*/}*.js'
      ],
      port: 9000,
      server: {
        baseDir: ['.tmp', config.app],
        routes: {
          '/bower_components': './bower_components'
        }
      }
    }
  },
```

When I change and save `app/index.html`, browserSync works well and reload the page. However, when I change and save `app/scripts/main.js` or `app/styles/main.css`.  The `watch:babel` or `watch:styles` tasks are triggered and the compiled `js` or `css` file are generated in `.tmp` folder, but somehow the browser doesn't reload at all...

gaboesquivel:
@silvenon we need set browserSync [watchOptions.ignored](http://www.browsersync.io/docs/options/#option-watchOptions) to prevent the js/css reload issues.
the problem was browserSync ignores files/folders that start with dot by default ( see https://github.com/BrowserSync/browser-sync/blob/master/lib/file-watcher.js#L15 )

fixes #585, #579 , #603

### DIFF
--- a/app/templates/Gruntfile.js
+++ b/app/templates/Gruntfile.js
@@ -68,7 +68,10 @@ module.exports = function (grunt) {
     browserSync: {
       options: {
         notify: false,
-        background: true
+        background: true,
+        watchOptions: {
+          ignored: ''
+        }
       },
       livereload: {
         options: {


silvenon:
Thanks so much! I was about to go for the raw browser-sync module solution, but this works just fine :smile: The `.css` files in the `app` directory don't get refreshed, thought, but one problem at a time.

Issue #614, opened 2015-10-28
apple-touch-icon.png is not optimized in imagemin


lachezar:
It appears that only files in `app/images` are configured to be optimized by the Gruntfile template and the apple-touch-icon is not processed by this task.

I will open a PR with fix for this soon.

sindresorhus:
Closing as this repo is deprecated.

PR #615, opened 2015-10-28
Run apple-touch-icon and favicon through the imagemin task

lachezar:
#614


addyosmani:
LGTM. We should have been doing this earlier.

PR #621, opened 2015-11-25
remove jshintrc file

neilhem:
Because of this project uses ESLint as linter there is no need to have this file in templates


silvenon:
Thanks :+1:

PR #626, opened 2015-12-14
Updated greetings log

DanielMSchmidt:
Hi there,

Just a small change, but as far as I can see jQuery is not included by default, so it's not out of the box in the generated app.
No big deal, but I wondered for a second, so I wanted to let you know ;)

Have a nice day!

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -58,7 +58,7 @@ module.exports = generators.Base.extend({
     var done = this.async();
 
     if (!this.options['skip-welcome-message']) {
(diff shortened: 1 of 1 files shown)


silvenon:
Hm, it should be included if Bootstrap is not present (as jQuery is a dependency of Bootstrap, so we don't need to add it explicitly).

When you don't select Bootstrap, the prompt asks you if you want to include jQuery, which is on by default, right?

DanielMSchmidt:
Yes, but it's not in there out of the box ;)

silvenon:
Oh, right, we made it optional at one point. Thanks.

Issue #567, opened 2015-07-06
Deprecation


silvenon:
I would like to deprecate this in favor of [generator-gulp-webapp](https://github.com/yeoman/generator-gulp-webapp). What do we need to do in order to achieve this?

If we want to rename (do we?):
- [ ] rename this to generator-grunt-webapp
- [ ] rename generator-gulp-webapp to generator-webapp
- [ ] `npm deprecate` generator-gulp-webapp (?)

Other tasks:
- [ ] update the docs with the deprecation notice

Questions:
- do we still do critical bugfixes?
- do we accept PRs with new features?

sindresorhus:
:+1: 

> do we still do critical bugfixes?
> do we accept PRs with new features?

As long as someone on the team is willing to review/merge/release, why not. I'm not interested however.

eddiemonge:
I actually want to take this in a different direction. It'd be nice if this one became a real base we could build (compose) on. I'm not sure if that means stripping out the build system and let other generators add that or something else. I just know it would be nice if I could require this in generator-angular or generator-backbone (is that still around?) or even generator-gulp-webapp and this would do the base site. Then the others could add the build system or more specific "recipes".

arthurvr:
> I actually want to take this in a different direction. It'd be nice if this one became a real base we could build (compose) on.

I like the idea but I wonder how realistic it is. Other build system require other modifications to the templates, every framework has it's own way of managing templates, other local server setups require other paths, ... That would all be things we'd need to work around.

> generator-backbone (is that still around?)

Yes, but it's unmaintained and all the things are broken. It's really sad and nobody seems willing to pick it up.

neilhem:
Would be great combining two build tools (generator-gulp-webapp & generator-webapp) on this project giving users choice between two build systems.

silvenon:
We already considered this in #565, but realized that we don't want to develop the Grunt version further.

alexconlin:
do you think there should be a deprecation warning in the readme? i'm starting a new project and just found this deprecation warning so am now switching to <a href="https://github.com/yeoman/generator-gulp-webapp">generator-gulp-webapp</a>. others might want to do the same?

silvenon:
Definitely, thanks for the suggestion.

gaboesquivel:
To deprecate why not simply transfer it to https://github.com/deprecate, https://help.github.com/articles/transferring-a-repository/

adrinux:
[...]

PR #627, opened 2015-12-17
Add deprecation notice

silvenon:
@sindresorhus sounds good?

I'm not sure about Yeoman's views on Grunt vs. gulp, I don't want to be too opinionated. Last time I checked it's this:

> The Yeoman team don't have any plans on dropping our support for Grunt at all. On the contrary, we feel that both Grunt and Gulp can happily co-exist and hope to support both communities with some automation tooling as best we can.

But that seems dated.

Refs #567.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,6 +2,12 @@
 
 [Yeoman](http://yeoman.io) generator that scaffolds out a front-end web app.
 
+## Looking for New Maintainers
+
+This generator is not actively maintained because we're focusing on [generator-gulp-webapp](https://github.com/yeoman/generator-gulp-webapp), which is very similar, but uses [gulp](http://gulpjs.com) instead of Grunt.
+
+We would be happy to take on new maintainers and will still be around for merging PRs.
+
 ![](http://i.imgur.com/uKTT2Hj.png)
 
 ## Features


sindresorhus:
@silvenon I think it would be better to say that it's unmaintained (not deprecated) and tell users we would be happy to take on new maintainers and merge PRs, but our focus is on generator-gulp-webapp.

silvenon:
Thanks for the feedback, updated.

silvenon:
The commit message still says "deprecation notice", my bad :sweat:

stevemao:
@silvenon @sindresorhus I'm happy to be around and review issues or prs. Since I'm quite familiar with the yeoman-generator code base..

stevemao:
@silvenon @sindresorhus How to apply for collaborator here?

silvenon:
@stevemao sorry for the late response. Not sure why, but even though I'm an owner I can't seem to find the repo settings. Maybe @sindresorhus can help out with adding you as a collaborator.

stevemao:
Thanks @silvenon :)

sindresorhus:
Done

stevemao:
Thanks @sindresorhus :)

Issue #632, opened 2016-01-15
Cannot grunt build app without running into errors


georaldc:
Was testing this to build a simple site and after installing and running the generator (which ran fine without any errors), grunt build would abort, probably due to the following:

```
Running "useminPrepare:html" (useminPrepare) task
Warning: An error occurred while processing a template (Unexpected token )). Use --force to continue.
```

georaldc:
Oh, I guess adding the --force flag makes it complete the build process, but it runs into another warning on the way:

```
Running "imagemin:dist" (imagemin) task
Warning: An error occurred while processing a template (Unexpected token )). Used --force, continuing.
```

stefanmeschke:
I think the problem is that the `useminPrepare:html` task is done after cleaning the `dist` directory and there is no html file to prepare. Same problem on the `imagemin:dist` task: no images the error occured?

terrencekeane:
I was getting the same error and found that the gulp version of this generator, generator-gulp-webapp, doesn't seem to have the error. I just switched over to using that generator for the project I was starting. Not a solution but figured it wouldn't hurt to mention it here.

tsck:
I believe the problem was that there were extra '%' characters added to the closing tags in the imagemin task.

Was:

``` json
cwd: '<%%= config.app %%>',
src: '*.{ico,png}',
dest: '<%%= config.dist %%>'
```

Fixed:

``` json
cwd: '<%%= config.app %>',
src: '*.{ico,png}',
dest: '<%%= config.dist %>'
```

I issued a pull request as per the above, just wanted to post it here for anyone having issues with this. Just delete those extra '%' characters and the build task should work.

RonanCodes:
Brilliant, it's working now.
The Yo Webapp setup still contains the error.

silvenon:
Released in [v1.1.2](https://github.com/yeoman/generator-webapp/releases/tag/v1.1.2).

PR #633, opened 2016-01-22
Fix errors on grunt build task

tsck:
On the grunt's build task, grunt was issuing the following errors:

```
Running "useminPrepare:html" (useminPrepare) task
Warning: An error occurred while processing a template (Unexpected token )). Use --force to continue.
```

and later:

```
Running "imagemin:dist" (imagemin) task
Warning: An error occurred while processing a template (Unexpected token )). Used --force, continuing.
```

The bug seems to have been caused by additional '%' characters added to the closing template
stings on the imagemin task in the gruntfile.

Resolves: #632


silvenon:
Thanks!

